Issue #52, opened 2019-06-21
Add metadata to Manual annotations


mtiberti:

mtiberti:
fixed in #159

PR #159, opened 2023-02-15
Addition of genomic_mutations (metadata) for manual annotations

katmeld:
Addresses issue #52. I have added the ability to parse an extra column (genomic_mutations) from the csv file for manual annotations and add the information to the position object. The code I have made is specific to genomic_mutations and is therefore not a complete fix of issue 52.

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -269,6 +269,7 @@ class cBioPortal(DynamicSource, object):
                 self.log.info("synonymous mutation %s discarded" % m)
                 continue
 
+            #Finds the mutation's indices in the complete list of mutations. Used for adding metadata
             mutation_indices = [i for i,x in enumerate(mutations) if x == m]
 
             try:
@@ -1545,7 +1546,7 @@ class MobiDB(DynamicSource):
 
 class ManualAnnotation(StaticSource):
 
-    _expected_cols = ['name', 'site', 'type', 'function', 'reference']
+    _expected_cols = ['name', 'site', 'type', 'function', 'reference','genomic_mutations']
     _ptm_keywords = ['ptm_cleavage', 'ptm_phosphorylation', 'ptm_ubiquitination', 'ptm_acetylation', 'ptm_sumoylation', 'ptm_nitrosylation', 'ptm_methylation']
     _supported_position_properties = _ptm_keywords
     _supported_sequence_properties = ['linear_motif', 'structure']
@@ -1601,10 +1602,25 @@ class ManualAnnotation(StaticSource):
         if self._df is None:
             return
 
+        #Prepares metadata for mutations
+        metadata.append('genomic_mutations')
+        gm_df = self._df['genomic_mutations'][self._df['type'] == 'mutation']
+        gm = list()
+        for row in gm_df:
+            if len(row.split(' ')) > 1:
+                md=str()
+                for el in row.split(' '):
+                    md+=el+','
+                gm.append(md[:-1].split(','))
+            elif row is not '':
+                gm.append(row.split(','))
+            else:
+                gm.append(None)
+        out_metadata = {'genomic_mutations':gm}
+
+        #mutation dataframe
         tmp_df = self._df[ self._df['type'] == 'mutation' ]
 
-        if 'genomic_mutations' in metadata:
-            out_metadata = { 'genomic_mutations' : tmp_df['function'].str.split(',').tolist() }
 
         mutations = tmp_df['site'].tolist()
 
@@ -1613,11 +1629,13 @@ class ManualAnnotation(StaticSource):
         self.log.info("unique mutations found in datafile: %s" % (", ".join(sorted(unique_mutations))))
 
         for m in unique_mutations:
+            #collects the position, wild type aa and mutated aa
             tokens = parse(self._mut_prot_parse, m)
             if tokens is None:
                 self.log.error(f"Failed to parse mutation {m}; check that the format is correct (i.e. HGVS protein single amino acid substitution)")
                 continue
 
+            #checks if it is a synonymous mutation
             num = int(tokens['position'])
             wt  = three_to_one(str.upper(tokens['wt']))
             mut = three_to_one(str.upper(tokens['mut']))
@@ -1626,6 +1644,7 @@ class ManualAnnotation(StaticSource):
                 self.log.info("synonymous mutation %s discarded" % m)
                 continue
 
+            #Checks if mutation is outside protein sequence
             try:
                 site_seq_idx = sequence.seq2index(num)
             except:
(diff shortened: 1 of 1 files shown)


mtiberti (changes requested):
great work!
Other than these comments, you should change the `docs/tutorial.md` file (which compiles into the tutorial on GitBook) to account for the new column and field (see the section on Manual annotations - and feel free to change what's written there now that doesn't conform with the current state of the code)

cancermuts/datasources.py
```diff
@@ -1601,10 +1602,25 @@ def add_mutations(self, sequence, metadata=[]):
...
+        metadata.append('genomic_mutations')
+        gm_df = self._df['genomic_mutations'][self._df['type'] == 'mutation']
+        gm = list()
+        for row in gm_df:
+            if len(row.split(' ')) > 1:
+                md=str()
+                for el in row.split(' '):
+                    md+=el+','
+                gm.append(md[:-1].split(','))
+            elif row is not '':
```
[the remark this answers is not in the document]
katmeld: it does not seem to give different output for an empty string

```
In [4]: pd.isna('')
Out[4]: False

In [5]: pd.isna('hg19,16:g.87435877A>G')
Out[5]: False
```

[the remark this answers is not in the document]
mtiberti: yes - my point was that `pandas` usually parses empty strings in a csv as `nan`, e.g.:

```
In [7]: import pandas as pd

In [8]: df = pd.read_csv('ma_test.csv', sep=';')

In [11]: print(df.genomic_mutations[1])
nan

In [12]: df.genomic_mutations[1] == ''
Out[12]: False

In [13]: pd.isna(df.genomic_mutations[1])
Out[13]: True
```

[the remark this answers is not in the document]
katmeld: Ahh okay, I get it. I'll have a look at it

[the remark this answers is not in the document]
katmeld:
```
In [4]: df = pd.read_csv('ma_test.csv',  keep_default_na=False)

In [5]: print(df)
                                                        name;site;type;function;reference;genomic_mutations
E25F;p.Glu25Phe;mutation;;;hg19   16:g.87435877A>G hg38                                   16:g.87402271A>G 
T50V;p.Thr50Val;mutation;;;                                                                                
K65V;p.Lys65Val;mutation;;;hg38   16:g.87402271A>G                                                         
E100F;p.Glu100Phe;mutation;;;hg19 16:g.87436693T>C                                                         
P94F;p.Pro94Phe;mutation;;;                                                                                
F13V;p.Phe13Val;mutation;;;                                                                                

In [6]: df = pd.read_csv('ma_test.csv')

In [7]: print(df)
                                                        name;site;type;function;reference;genomic_mutations
E25F;p.Glu25Phe;mutation;;;hg19   16:g.87435877A>G hg38                                   16:g.87402271A>G 
T50V;p.Thr50Val;mutation;;;       NaN                                                                  NaN 
K65V;p.Lys65Val;mutation;;;hg38   16:g.87402271A>G                                                     NaN 
E100F;p.Glu100Phe;mutation;;;hg19 16:g.87436693T>C                                                     NaN 
P94F;p.Pro94Phe;mutation;;;       NaN                                                                  NaN 
F13V;p.Phe13Val;mutation;;;       NaN                                                                  NaN 
```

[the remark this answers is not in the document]
mtiberti: got it! Sorry, I actually had a look at the parsing but missed that option.

I think we can leave things as they are, makes sense now :)

docs/tutorial.md
```diff
@@ -622,9 +623,11 @@ The column format changes depending on the `type`:
     * `name` can be any string
     * `site` needs to be a HGVS-format protein variant specification, e.g.
      `p.Ala398Tyr`
-    * `function` should be either empty or a HGVS-format single-nucleotide
+    * `genomic_mutations` should be either empty or a HGVS-format single-nucleotide
```
mtiberti, line 626: yep, but in all cases the genome build should be specified as `hg19` or `hg38`, not just `19` or `38` (here and in all mentions below)

docs/tutorial.md
```diff
@@ -614,6 +614,7 @@ information. The file should have the following columns, separated by `;`:
 | `type` | type of this feature; see below |
 | `function` | function description of this feature or functional annotation; see below |
 | `reference` | reference to the literature for this feature (if any) |
+| `genomic_mutations` | metadata for mutations; see below |
```
mtiberti, line 617: does the column need to be present even if we don't e.g. annotate mutations, or we don't use metadata? We should verify this and specify it in the docs. I think that if possible we shouldn't check for it if we're not using the metadata

docs/tutorial.md
```diff
@@ -668,6 +671,9 @@ using the CSV file works as you would expect:
 # adds mutations to the seq object
 >>> ma.add_mutations(seq)
 
+# adds mutations and the corresponding metadata to the seq obejct
```
mtiberti, line 674: yep - but can we have this example as a comment rather than an actually run line?

I'm asking this because the tutorial is constructed so that one can mindlessly paste the lines from it and run it. If we leave two of these lines we will annotate the mutations twice

cancermuts/datasources.py
```diff
@@ -1601,10 +1602,24 @@ def add_mutations(self, sequence, metadata=[]):
...
+        #Prepares metadata for mutations
+        do_genomic_mutations=False
+        if 'genomic_mutations' in metadata:
+            do_genomic_mutations=True
+
+        if do_genomic_mutations:
+            gm_df = self._df['genomic_mutations'][self._df['type'] == 'mutation']
+            gm = list()
+            for row in gm_df:
+                if row is not '':
```
mtiberti, line 1614: I realised that this should use the equality test operator `==` rather than `is`, which is used for object equality. Python also warns us:

```
In [1]: a = ''

In [2]: a is ''
<>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
<ipython-input-2-defaa20e97c9>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
  a is ''
Out[2]: True
```

mtiberti (changes requested):

docs/tutorial.md
katmeld: As it is now, the file is parsed by itself, and the metadata is only specified when adding mutations. To not parse the genomic_mutations column the user would need to specify the metadata when parsing the file, I think. Can you think of another way?

docs/tutorial.md
katmeld: Yes, in relation to this, is it alright if I remove >>> from all the code pieces? It would make it much easier to copy.

docs/tutorial.md
katmeld: Maybe we can parse the file without the column, and if the genomic_mutations is specified in metadata, then we can parse the column in add_mutations

mtiberti (approved):

PR #162, opened 2023-03-10
Addition of popmax allele freq. to gnomad annotations

### LINKED ISSUE #146: We should add popmax annotation to gnomAD

katmeld:
fixes #146. Adds popmax allele frequencies to the mutation object in the same way that allele frequencies are added.

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -1,6 +1,9 @@
 # datasources.py - data sources handling for the cancermuts package
 # (c) 2018 Matteo Tiberti <matteo.tiberti@gmail.com>
+# (c) 2023 Katrine Meldgård <katrine@meldgaard.dk>
 # This file is part of cancermuts
+# The function '_get_popmax_af' is taken and modified from the 'gnomad2csv' script
+# which is part of the ELELAB/CSB-scripts repository
 #
 # cancermuts is free software: you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
@@ -1249,7 +1252,7 @@ class gnomAD(DynamicSource, object):
                     add_this_metadata(mut, gene_id)
 
     def _get_exome_allele_freq(self, mutation, gene_id):
-        self.log.info("getting genome allele frequency")
+        self.log.info("getting exome allele frequency")
         self._get_metadata(mutation, gene_id, 'gnomad_exome_allele_frequency')
 
     def _get_genome_allele_freq(self, mutation, gene_id):
@@ -1257,10 +1260,30 @@ class gnomAD(DynamicSource, object):
         self._get_metadata(mutation, gene_id, 'gnomad_genome_allele_frequency')
 
     def _get_popmax_exome_allele_freq(self, mutation, gene_id):
-        self.log.info("getting popmax genome allele frequency")
+        """Adds popmax exome allele frequency to mutation object.
+
+        Parameters
+        ----------
+        mutation : :obj:'cancermuts.core.Mutation'
+            Object containing mutations and metadata associated with the sequence.
+        gene_id : 'str'
+            ID of the gene (gene name) to which the sequence belongs
+        """
+
+        self.log.info("getting popmax exome allele frequency")
         self._get_metadata(mutation, gene_id, 'gnomad_popmax_exome_allele_frequency')
 
     def _get_popmax_genome_allele_freq(self, mutation, gene_id):
+        """Adds popmax genome allele frequency to mutation object.
+
+        Parameters
+        ----------
+        mutation : :obj:'cancermuts.core.Mutation'
+            Object containing mutations and metadata associated with the sequence.
+        gene_id : 'str'
+            ID of the gene (gene name) to which the sequence belongs
+        """
+
         self.log.info("getting popmax genome allele frequency")
         self._get_metadata(mutation, gene_id, 'gnomad_popmax_genome_allele_frequency')
 
@@ -1332,9 +1355,9 @@ class gnomAD(DynamicSource, object):
                 self.log.warning("more than one entry for %s! Skipping" % v_str)
                 af = None
             elif len(this_df) == 1:
-                if np.isnan(this_df[exac_key[md_type]].values[0]):
+                if pd.isna(this_df[exac_key[md_type]].values[0]):
                     af = None
-                    mutation.metadata[md_type] = np.nan
+                    mutation.metadata[md_type] = []
                     self.log.info("nan entry found for %s" % v_str)
                 else:
                     af = this_df[exac_key[md_type]].values[0]
@@ -1391,7 +1414,7 @@ class gnomAD(DynamicSource, object):
                                                "refBuild"  : reference_genome,
                                                "dataset"   : dataset }
                                 }),
-                headers={"Content-Type": "application/json"}) 
+                headers={"Content-Type": "application/json"})
         except:
             self.log.error("Couldn't perform request for gnomAD")
             return None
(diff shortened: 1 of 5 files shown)


cancermuts/datasources.py
```diff
@@ -1322,6 +1334,7 @@ def _get_metadata(self, mutation, gene_id, md_type):
             elif len(this_df) == 1:
                 if np.isnan(this_df[exac_key[md_type]].values[0]):
                     af = None
+                    mutation.metadata[md_type] = np.nan
```
mtiberti, line 1337: actually this should be `pd.NA` which is more appropriate for use in `pandas`

cancermuts/datasources.py
```diff
@@ -1368,7 +1391,7 @@ def _get_gnomad_data(self, gene_id, reference_genome, dataset):
                                                "refBuild"  : reference_genome,
                                                "dataset"   : dataset }
                                 }),
-                headers={"Content-Type": "application/json"})
+                headers={"Content-Type": "application/json"})
```
mtiberti, line 1394: trailing space to be removed

cancermuts/datasources.py
```diff
@@ -1404,8 +1427,106 @@ def _get_gnomad_data(self, gene_id, reference_genome, dataset):
...
+            popmax_tot_ac = pd.NA
+            popmax_tot_an = pd.NA
+            popmax_tot_af = pd.NA
+
+        return pd.Series([popmax_exome_ac,  popmax_exome_an,  popmax_exome_af,
+                        popmax_genome_ac, popmax_genome_an, popmax_genome_af,
+                        popmax_tot_ac,    popmax_tot_an,    popmax_tot_af]).replace(pd.NA,np.nan)
+
+
+
```
mtiberti, line 1529: extra newlines to be removed

cancermuts/datasources.py
```diff
@@ -1404,8 +1427,106 @@ def _get_gnomad_data(self, gene_id, reference_genome, dataset):
         variants['genome_af'] = variants['genome_ac'] / variants['genome_an']
         variants['total_af'] = variants['total_ac'] / variants['total_an']
 
+        variants[['popmax_exome_ac',  'popmax_exome_an',  'popmax_exome_af',
+              'popmax_genome_ac', 'popmax_genome_an', 'popmax_genome_af',
+              'popmax_tot_ac',    'popmax_tot_an',    'popmax_tot_af']] = variants.apply(self.get_popmax_af, axis=1)
+
         return variants
 
+    def get_popmax_af(self, variants, pops=['afr', 'eas', 'nfe', 'amr', 'sas']):
```
mtiberti, line 1436: can we
* change the name of the method according to the "private method" naming convention (so it should start with `_`)
* At the beginning of the file, under the copyright section, add where you took this from - this is necessary because the code is licensed under GPL
* add yourself at the beginning in the copyright section, with a line similar to mine (`(c) 2018 Matteo Tiberti <matteo.tiberti@gmail.com>`). We should remember to do this in every file you modify so you're properly credited :) I forgot to point it out in the previous PRs, apologies! We can add it retroactively

cancermuts/datasources.py
```diff
@@ -1254,6 +1256,14 @@ def _get_genome_allele_freq(self, mutation, gene_id):
         self.log.info("getting genome allele frequency")
         self._get_metadata(mutation, gene_id, 'gnomad_genome_allele_frequency')
 
+    def _get_popmax_exome_allele_freq(self, mutation, gene_id):
```
mtiberti, line 1259: can we add a docstring to new functions that we make?

I plan eventually on having the whole code documented through docstrings. I had started doing it e.g. in `core.py` or in here - then I had to stop with lack of time, but it would be good to have docs for the functions that we are adding from scratch at least. 

You can find an example of the format in `core.py` - I tried using the [`numpy` docstring style](https://numpydoc.readthedocs.io/en/latest/format.html) which is popular in scientific Python packages.

mtiberti (changes requested):

cancermuts/datasources.py
```diff
@@ -1194,19 +1203,32 @@ def __init__(self, version='2.1'):
...
         gnomADGenomeAlleleFrequency.set_version_in_desc(self._version_str[version])
+        gnomADPopmaxExomeAlleleFrequency.set_version_in_desc(self._version_str[version])
+        gnomADPopmaxGenomeAlleleFrequency.set_version_in_desc(self._version_str[version])
 
     def add_metadata(self, sequence, md_type=['gnomad_exome_allele_frequency'], use_alias=None):
-        
+        if '2.1' in self._gnomad_version:
+            for md in md_type:
+                if md not in self._version_metadata_compatability['2.1']:
+                    self.log.error(f"The  metadata type '{md}' is not compatible with the gnomAD version. Compatible metadata are {self._version_metadata_compatability['2.1']}")
```
mtiberti, line 1226: yep - we should `raise` an exception here - For example `ValueError`

cancermuts/datasources.py
```diff
@@ -1194,19 +1203,32 @@ def __init__(self, version='2.1'):
...
+        elif '3' in self._gnomad_version:
+            for md in md_type:
+                if md not in self._version_metadata_compatability['3']:
+                    self.log.error(f"The metadata type '{md}' is not compatible with the gnomAD version. Compatible metadata are {self._version_metadata_compatability['3']}")
```
mtiberti, line 1230: same as before

cancermuts/datasources.py
```diff
@@ -1189,6 +1189,12 @@ class gnomAD(DynamicSource, object):
                        'exac' :            'EXaC',
                     }
 
+    _version_metadata_compatability = { '2.1' :         ['gnomad_exome_allele_frequency',
```
mtiberti, line 1192: here I think we want to add the various `2.1_controls`, `2.1_non-neuro` etc.

cancermuts/datasources.py
```diff
@@ -1209,9 +1216,19 @@ def __init__(self, version='2.1'):
 
         gnomADExomeAlleleFrequency.set_version_in_desc(self._version_str[version])
         gnomADGenomeAlleleFrequency.set_version_in_desc(self._version_str[version])
+        gnomADPopmaxExomeAlleleFrequency.set_version_in_desc(self._version_str[version])
+        gnomADPopmaxGenomeAlleleFrequency.set_version_in_desc(self._version_str[version])
 
     def add_metadata(self, sequence, md_type=['gnomad_exome_allele_frequency'], use_alias=None):
-        
+        if '2.1' in self._gnomad_version:
```
mtiberti, line 1223: I think we should rely on `self._version_metadata_compatability` to make this decision, instead of hardcoding checks on `2.1` or `3`, e.g.:

```
for md in md_type:
    if not md in self._version_metadata_compatability[self._gnomad_version]:
        raise ValueError(...)
```

if we keep things parametric we don't need to do anything else when we e.g. add new versions of gnomAD in the future (adding version 4 would just be a matter of adding some appropriate entries in the dictionaries above)

cancermuts/datasources.py
```diff
@@ -1443,16 +1460,27 @@ def _get_gnomad_data(self, gene_id, reference_genome, dataset):
 
         variants['edited_variant_id'] = variants.apply(self._edit_variant_id, axis=1)
 
-        variants['total_ac'] = variants['exome_ac'] + variants['genome_ac']
-        variants['total_an'] = variants['exome_an'] + variants['genome_an']
+        if self._gnomad_version!='3':
```
mtiberti, line 1463: again, can we make this decision parametric? See example in line 1223

mtiberti (changes requested):

cancermuts/datasources.py
```diff
@@ -1200,13 +1218,21 @@ def __init__(self, version='2.1'):
...
         gnomADGenomeAlleleFrequency.set_version_in_desc(self._version_str[version])
+        gnomADPopmaxExomeAlleleFrequency.set_version_in_desc(self._version_str[version])
+        gnomADPopmaxGenomeAlleleFrequency.set_version_in_desc(self._version_str[version])
 
     def add_metadata(self, sequence, md_type=['gnomad_exome_allele_frequency'], use_alias=None):
-        
+        for md in md_type:
+            if md not in self._version_metadata_compatability[self._gnomad_version]:
+                self.log.error(f"The  metadata type '{md}' is not compatible with the gnomAD version. Compatible metadata are {self._version_metadata_compatability[self._gnomad_version]}")
+                raise ValueError
```
mtiberti, line 1234: can we also have the error as exception message? So it's easier to debug for who might encounter it

mtiberti (changes requested):

mtiberti (approved):

PR #172, opened 2023-04-27
Check for cosmic database file

### LINKED ISSUE #169: COSMIC() crashes when no file is specified
In this case, I think the default path is just outdated, but it would be nice to not crash like this

```Traceback (most recent call last):
  File "./test.py", line 8, in <module>
    cs=COSMIC()
  File "/data/user/kame/devel/cancermuts/cancermuts/cancermuts/log.py", line 50, in wrapper
    function(*args, **kwargs)
  File "/data/user/kame/devel/cancermuts/cancermuts/cancermuts/datasources.py", line 546, in __init__
    dataframes.append( pd.read_csv(f, sep='\t', dtype='str', na_values='NS', usecols=self._use_cols_snp, encoding=encodings[fi]) )
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/parsers/readers.py", line 586, in read_csv
    return _read(filepath_or_buffer, kwds)
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/parsers/readers.py", line 482, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/parsers/readers.py", line 811, in __init__
    self._engine = self._make_engine(self.engine)
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/parsers/readers.py", line 1040, in _make_engine
    return mapping[engine](self.f, **self.options)  # type: ignore[call-arg]
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/parsers/c_parser_wrapper.py", line 51, in __init__
    self._open_handles(src, kwds)
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/parsers/base_parser.py", line 229, in _open_handles
    errors=kwds.get("encoding_errors", "strict"),
  File "/data/user/kame/devel/cancermuts/cancermuts_env/lib/python3.7/site-packages/pandas/io/common.py", line 707, in get_handle
    newline="",

FileNotFoundError: [Errno 2] No such file or directory: '/data/databases/cosmic/CosmicMutantExport.tsv'

### PR DESCRIPTION
**Pull request by katmeld**
fixes #169

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -534,11 +534,8 @@ class COSMIC(DynamicSource, object):
         dataframes = []
 
         if database_files is None:
-            database_dir   = '/data/databases/cosmic'
-            databases      = [  'CosmicMutantExport' ]
-            self._database_files = [os.path.join(database_dir, i)+'.tsv' for i in databases]
-        else:
-            self._database_files = database_files
+            self.log.error('The database file for COSMIC is not specified.')
+            raise FileNotFoundError('The database file for COSMIC is not specified.')
 
         if database_encoding is None or isinstance(database_encoding, str):
             encodings = [ database_encoding for i in self._database_files ]
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -534,9 +534,12 @@ class COSMIC(DynamicSource, object):
         dataframes = []
 
         if isinstance(database_files, str):
-            self._database_files = list(database_files)
+            self._database_files = [database_files]
         else:
-            self._database_files = database_files
+            if not hasattr(database_files,'__iter__'):
+                self.log.error('COSMIC database_files does not have the correct format.')
+                raise TypeError('COSMIC database_files does not have the correct format.')
+        self._database_files = database_files
 
         if database_encoding is None or isinstance(database_encoding, str):
             encodings = [ database_encoding for i in self._database_files ]
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -504,7 +504,7 @@ class cBioPortal(DynamicSource, object):
 
 class COSMIC(DynamicSource, object):
     @logger_init
-    def __init__(self, database_files=None, database_encoding=None):
+    def __init__(self, database_files, database_encoding=None):
         description = "COSMIC Database"
         super(COSMIC, self).__init__(name='COSMIC', version='v87', description=description)
 
@@ -533,9 +533,10 @@ class COSMIC(DynamicSource, object):
 
         dataframes = []
 
-        if database_files is None:
-            self.log.error('The database file for COSMIC is not specified.')
-            raise FileNotFoundError('The database file for COSMIC is not specified.')
+        if isinstance(database_files, str):
+            self._database_files = list(database_files)
+        else:
+            self._database_files = database_files
 
         if database_encoding is None or isinstance(database_encoding, str):
             encodings = [ database_encoding for i in self._database_files ]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cancermuts/datasources.py [hunk 1]
```diff
@@ -534,11 +534,8 @@ def __init__(self, database_files=None, database_encoding=None):
         dataframes = []
 
         if database_files is None:
-            database_dir   = '/data/databases/cosmic'
-            databases      = [  'CosmicMutantExport' ]
-            self._database_files = [os.path.join(database_dir, i)+'.tsv' for i in databases]
-        else:
-            self._database_files = database_files
+            self.log.error('The database file for COSMIC is not specified.')
+            raise FileNotFoundError('The database file for COSMIC is not specified.')
```
mtiberti, line 538: now that I see this, I think it would be a better idea to just make the `database_files` argument mandatory (i.e. without a default argument). This way we wouldn't even have to check and would just use a basic feature of the language :)). I think we should turn this check into checking if there is at least one element in the list

mtiberti (changes requested):

cancermuts/datasources.py
```diff
@@ -533,10 +533,8 @@ def __init__(self, database_files=None, database_encoding=None):
 
         dataframes = []
 
-        if database_files is None:
-            database_dir   = '/data/databases/cosmic'
-            databases      = [  'CosmicMutantExport' ]
-            self._database_files = [os.path.join(database_dir, i)+'.tsv' for i in databases]
+        if isinstance(database_files, str):
+            self._database_files = list(database_files)
```
mtiberti, line 537: does this not become a list of characters? e.g.

```
In [2]: list("asd")
Out[2]: ['a', 's', 'd']
```

probably you want like

```
self._database_files = [database_files]
```

cancermuts/datasources.py
```diff
@@ -533,10 +533,8 @@ def __init__(self, database_files=None, database_encoding=None):
...
         else:
             self._database_files = database_files
```
mtiberti, line 539: also add check on the fact that it is an interable with e.g. `hasattr(database_files, '__iter__')` as we discussed yesterday and raise an exception if it's not (the argument should be either a string or an iterable)

mtiberti (changes requested):

cancermuts/datasources.py
```diff
@@ -533,12 +533,13 @@ def __init__(self, database_files=None, database_encoding=None):
...
-            databases      = [  'CosmicMutantExport' ]
-            self._database_files = [os.path.join(database_dir, i)+'.tsv' for i in databases]
+        if isinstance(database_files, str):
+            self._database_files = [database_files]
         else:
-            self._database_files = database_files
+            if not hasattr(database_files,'__iter__'):
+                self.log.error('COSMIC database_files does not have the correct format.')
+                raise TypeError('COSMIC database_files does not have the correct format.')
+        self._database_files = database_files
```
mtiberti, line 542: in my testing, this leads to a crash when `database_files` is a string - that's because it's re-assigned here even when it is first assigned after the `if` above (I think you just need to put this statement in the `else` code block)

mtiberti (approved):

PR #179, opened 2023-06-22
Adds handling of all-nan cases for `popmax` calculation in gnomAD

mtiberti:


mtiberti:
tested by @SimoneScrima

PR #206, opened 2025-03-14
Fix handling of multiple entrez IDs, log warning

### LINKED ISSUE #205: `datasources.UniProt.get_alisases` has unhandled `raise` when multiple Entrez ID are found
traceback:

```
"/data/user/shared_projects/mavisp_automatization_collection/EiriniGian/mavisp_templates/GENE_NAME/cancermuts_metatable/pancancer.py", line 24, in <module>
    seq = up.get_sequence(args.prt, upid=args.uniprotID, upac=args.uniprotAC)
  File "/usr/local/envs/cancermuts/lib/python3.8/site-packages/cancermuts/datasources.py", line 158, in get_sequence
    this_entrez = self._get_aliases(this_upac, ['GeneID'])
  File "/usr/local/envs/cancermuts/lib/python3.8/site-packages/cancermuts/datasources.py", line 236, in _get_aliases
    raise TypeError
TypeError
```

this happens in cases in which multiple Gene ID (=multiple genes) are found for the same UniProt ID. It should probably handle the problem by not assigning an Entrez ID at all, or multiple of them.

EszterToldi:
fixes #205

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -232,8 +232,10 @@ class UniProt(DynamicSource, object):
             if len(responses) == 0:
                 self.log.warning(f"No {t} found for {gene_id}")
                 return None
+
             if len(responses) > 1:
-                raise TypeError
+            	self.log.warning(f"Multiple {t} found for {gene_id}. Entrez ID processing and mutation retrieval from cBioPortal will be skipped. Found {t}: {[response['to'] for response in responses]}")
+            	return None
 
             results = responses[0]
 
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -234,8 +234,9 @@ class UniProt(DynamicSource, object):
                 return None
 
             if len(responses) > 1:
-            	self.log.warning(f"Multiple {t} found for {gene_id}. Entrez ID processing and mutation retrieval from cBioPortal will be skipped. Found {t}: {[response['to'] for response in responses]}")
-            	return None
+                self.log.warning(f"Multiple {t} found for {gene_id}. Found {t}: {', '.join(response['to'] for response in responses)}. No {t} will be assigned.")
+                out[t] = None
+                continue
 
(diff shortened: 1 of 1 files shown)


cancermuts/datasources.py
```diff
@@ -232,8 +232,10 @@ def _get_aliases(self, gene_id, to, fr='UniProtKB_AC-ID'):
             if len(responses) == 0:
                 self.log.warning(f"No {t} found for {gene_id}")
                 return None
+
             if len(responses) > 1:
-                raise TypeError
+            	self.log.warning(f"Multiple {t} found for {gene_id}. Entrez ID processing and mutation retrieval from cBioPortal will be skipped. Found {t}: {[response['to'] for response in responses]}")
```
mtiberti, line 237: can you please
* remove the sentence that says that cBioPortal will be skipped, since this is not in the right place for it
* can the responses be comma-separated in the warning? e.g. `asd, qwe`. You can `', '.join()` for that
* can you also state that no identifier will be assigned for that type? (e.g. no Entrez ID will be assigned)

cancermuts/datasources.py
```diff
@@ -232,8 +232,10 @@ def _get_aliases(self, gene_id, to, fr='UniProtKB_AC-ID'):
...
+            	return None
```
mtiberti, line 238: I don't think just returning `None` here works; normally we return a dictionary of responses, one per `t` in the loop. I would just set that one to `None`, and let the others as they are since they might just be valid, and then return the `out` dictionary as we already do

mtiberti (changes requested):

EszterToldi:
I now removed the sentence about cBioPortal, made the responses in the warning comma-separated, and added a sentence that no identifier will be assigned. I also set t to none, so that it should return the out dictionary as you requested. Please check if further changes are needed

mtiberti (changes requested):

PR #214, opened 2025-05-20
Use Biothings to interrogate Myvariant

### LINKED ISSUE #55: Use biothings to interrogate MyVariant
see https://biothings-clientpy.readthedocs.io/en/latest/doc/quickstart.html#use-the-client-for-myvariant-info-api-variants

EszterToldi:
fixes #55

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -43,6 +43,7 @@ import re
 import sys
 import os
 import myvariant
+from biothings_client import get_client
 import pyliftover
 import itertools
 from parse import parse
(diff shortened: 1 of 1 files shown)


cancermuts/datasources.py
```diff
@@ -43,6 +43,7 @@
 import sys
 import os
 import myvariant
```
mtiberti, line 45: we don't need this import anymore

please also check that

  - requirements in `setup.py` are still consistent with the code as it is now (which I don't think they are right now). We need the package to still install properly
  - if we need any change in the documentation following this

mtiberti (changes requested):
can you please stash/remove all change that don't have to do with the current issue?

then we can add them in as another PR

mtiberti (approved):

PR #216, opened 2025-05-22
Shorten MyVariant code

### LINKED ISSUE #215: Shorten and simplify MyVariant code

EszterToldi:
fixes #215

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -42,7 +42,6 @@ import json
 import re
 import sys
 import os
-import myvariant
 from biothings_client import get_client
 import pyliftover
 import itertools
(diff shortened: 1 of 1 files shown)


cancermuts/datasources.py
```diff
@@ -907,96 +907,92 @@ def _get_revel(self, mutation):
...
 
-        mutation.metadata['revel_score'] = list()
+        mutation.metadata['revel_score'] = []
 
-        if 'genomic_coordinates' not in mutation.metadata and 'genomic_mutations' not in mutation.metadata:
-            self.log.warning("no genomic coordinates or genomic mutations data for mutation %s; it will be skipped" % mutation)
+        gcs = mutation.metadata.get('genomic_coordinates', [])
+        gms = mutation.metadata.get('genomic_mutations', [])
+
+        if (gcs is None or gcs == [] or gcs == '') and (gms is None or gms == [] or gms == ''):
```
mtiberti, line 915: why do we check for empty strings? Do we have cases in which this is actually happening? I was wondering because I don't think we should ever have that, if we have that it might be an issue

mtiberti (changes requested):

mtiberti (approved):

Issue #208, opened 2025-05-20
change `datasources.UniProt` to support isoforms different from the canonical one


mtiberti:

mtiberti:
done in #221

Issue #43, opened 2019-04-17
Better support for multiple isoforms


mtiberti:
Right now it only supports the canonical Uniprot one. This can be a problem when it's not available for instance in COSMIC. It can also be a problem with misannotated entries in cBioPortal (as it happened with AMBRA1 in the past)

mtiberti:
all done! 👏

PR #221, opened 2025-07-07
Adding uniprot_AC and isoform tracking to core.py (for isoforms support)

pablosanchezizquierdo:
fix sub issue #207, and #208 from #43  adding support for isoforms in the ` core.py` and `datasources.py`

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -120,9 +120,9 @@ the analysis and links together all the information we are going to collect. In 
 create it we use a data source class for UniProt from which we can downlaod protein sequences.
 This is only source for sequences available at the moment, more can be added in the future.
 
-To download the main UniProt sequence, you will need the corresponding HGCN symbol only.
-The UniProt object automatically identifies the most likely UniProt ID corresponding to the
-gene name. Alternatively, it is possible to provide the UniProt ID manually as an option.
+To download the canonical UniProt sequence, you will need the corresponding HGCN symbol only.
+The UniProt object automatically identifies the most likely UniProt ID and AC corresponding to the
+gene name. Alternatively, it is possible to provide the UniProt ID manually.
 
 This is done as follows:
 
@@ -137,8 +137,8 @@ This is done as follows:
 # get the sequence for the protein
 >>> seq = up.get_sequence('MAP1LC3B')
 
-# alternatively, we can specifically ask for a Uniprot ID
->>> seq = up.get_sequence('MAP1LC3B', upid='MLP3B_HUMAN')
+# alternatively, we can specifically ask for a Uniprot ID and/or AC
+>>> seq = up.get_sequence('MAP1LC3B', upid='MLP3B_HUMAN', upac='Q9GZQ8')
 
 # this prints the downloaded protein sequence
 >>> print(seq.sequence)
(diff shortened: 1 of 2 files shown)


pablosanchezizquierdo:
do we want to make uniprot_AC mandatory?

cancermuts/core.py
```diff
@@ -124,8 +130,10 @@ def __iter__(self):
         return iter(self.positions)
 
     def __repr__(self):
-        return "<Sequence of %s from %s, %d positions>" % (self.gene_id, self.source.name, len(self.positions))
-
+        form = "canonical" if self.is_canonical else "non_canonical"
+        iso_str = self.isoform if self.isoform else "no_isoform"
+        return f"<Sequence {self.gene_id} ({form}, isoform={iso_str}, uniprot_ac={self.uniprot_ac}) from {self.source.name}, {len(self.positions)} positions>"
```
mtiberti, line 135: since this is the textual representation of an object, I think it's just fair to print the state of its attributes, e.g.:

```
<Sequence {self.gene_id} (uniprot_ac=..., isoform=..., is_canonical=..., from {self.source.name}, {len(self.positions)} positions>
```

cancermuts/datasources.py
```diff
@@ -121,8 +121,12 @@ def __init__(self, *args, **kwargs):
         super(UniProt, self).__init__(name='UniProt', version='1.0', description=description)
         self._uniprot_service = bsUniProt()
 
-    def get_sequence(self, gene_id, upid=None, upac=None):
-        if upac is not None or upid is not None:
+    def get_sequence(self, gene_id, upid=None, upac=None, isoform=None, is_canonical=True):
```
mtiberti, line 124: I think we need to restructure this a little to be sure we are consistent, especially because `Sequence` is kinda dumb, and so we need to a bit more smarts here :) I think it would make sense to first infer the correct UniProt AC if it's not provided, and then work on the isoform. In particular:

  - if the isoform is supplied, then upac at least should also be supplied. This is because we want to avoid cases in which the user provides an isoform, but we select a different upac. Therefore, cases in which `upac` is not selected will always use the canonical isoform
  - once we selected the upac (or used the provided one), we should try and find the isoform corresponding to canonical (if possible) and use that to annotate the `Sequence` object. If the isoform is provided, we should also check that it matches the upac and that it exists

cancermuts/datasources.py
```diff
@@ -153,21 +153,30 @@ def get_sequence(self, gene_id, upid=None, upac=None, isoform=None):
...
+            for prod in alt_prods:
+                for iso in prod.get("isoforms", []):
+                    ids = iso.get("isoformIds", [])
+                    status = iso.get("isoformSequenceStatus", "")
+                    if isoform in ids:
+                        isoform_found = True
+                        is_canonical = (status == "Displayed")
+                        break
+                if isoform_found:
+                    break
```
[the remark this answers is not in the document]
pablosanchezizquierdo: No, you are right I think we can directly use []

cancermuts/datasources.py
```diff
@@ -159,13 +159,13 @@ def get_sequence(self, gene_id, upid=None, upac=None, isoform=None):
...
             for prod in alt_prods:
-                for iso in prod.get("isoforms", []):
-                    ids = iso.get("isoformIds", [])
-                    status = iso.get("isoformSequenceStatus", "")
+                for iso in prod["isoforms"]:
+                    ids = iso["isoformIds"]
+                    status = iso["isoformSequenceStatus"]
                     if isoform in ids:
                         isoform_found = True
                         is_canonical = (status == "Displayed")
```
mtiberti, line 171: this is better, but we should handle `KeyError`s or other problems if they happen, e.g.:

```
alt_prods = [x for x in data["comments"] if x["commentType"] == "ALTERNATIVE PRODUCTS"]
if len(alt_prods) == 0: 
    raise ValueError("...:)

try:
    ids = iso[...]
except KeyError:
    raise ValueError(...)
```

mtiberti (changes requested):

PR #222, opened 2025-07-21
Adding safeguard for cBioPortal when non-canonical isoform is used

### LINKED ISSUE #43: Better support for multiple isoforms

pablosanchezizquierdo:
fixes #43 subissue #209

Updated tutorial and datasources.py with a safeguard when using non-canonical isoform

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -336,12 +336,12 @@ class cBioPortal(DynamicSource, object):
         _cBioPortal_supported_metadata = ['cancer_type', 'cancer_study', 'genomic_coordinates', 'genomic_mutations']
 
         if not sequence.is_canonical:
-            raise UnexpectedIsoformError("cBioPortal mutation annotation only supports canonical isoforms. Please use a canonical UniProt sequence")
+            raise UnexpectedIsoformError("cBioPortal mutation annotation only supports canonical isoforms. Please use a Sequence object for a canonical isoform")
 
         if 'entrez' not in sequence.aliases.keys() or sequence.aliases['entrez'] is None:
             self.log.error('Entrez ID alias not available in sequence object')
             raise TypeError('Entrez ID alias not available in sequence object')
-        
+
         for md in metadata:
             if md not in _cBioPortal_supported_metadata:
                 self.log.error(f'{md} is not a valid metadata. Supported metadata are: {_cBioPortal_supported_metadata}')
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -337,6 +337,9 @@ class cBioPortal(DynamicSource, object):
         if 'entrez' not in sequence.aliases.keys() or sequence.aliases['entrez'] is None:
             self.log.error('Entrez ID alias not available in sequence object')
             raise TypeError('Entrez ID alias not available in sequence object')
+        
+        if hasattr(sequence, "is_canonical") and not sequence.is_canonical:
+            raise ValueError("cBioPortal mutation annotation only supports canonical isoforms.")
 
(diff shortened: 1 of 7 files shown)


cancermuts/datasources.py
```diff
@@ -337,6 +338,9 @@ def add_mutations(self, sequence, metadata=[]):
         if 'entrez' not in sequence.aliases.keys() or sequence.aliases['entrez'] is None:
             self.log.error('Entrez ID alias not available in sequence object')
             raise TypeError('Entrez ID alias not available in sequence object')
+        
+        if hasattr(sequence, "is_canonical") and not sequence.is_canonical:
```
mtiberti, line 342: are there cases in which an instance of `cancermuts.core.Sequence` doesn't have `is_canonical` set?

in principle there shouldn't be

mtiberti (commented):
- reading this, I realized it would be just better to have a tutorial for and using a non-canonical isoform, can you prepare it?
  - can you also add an example of the behaviour in case it finds a non-canonical isoform in the docs?

cancermuts/datasources.py
```diff
@@ -337,6 +338,9 @@ def add_mutations(self, sequence, metadata=[]):
         if 'entrez' not in sequence.aliases.keys() or sequence.aliases['entrez'] is None:
             self.log.error('Entrez ID alias not available in sequence object')
             raise TypeError('Entrez ID alias not available in sequence object')
+        
+        if not sequence.is_canonical:
+            raise UnexpectedIsoformError("cBioPortal mutation annotation only supports canonical isoforms. Please use a canonical UniProt sequence")
```
mtiberti, line 343: I think this should be the very first thing we check for, since if this is the case nothing can be done

docs/tutorial_isoforms.py
mtiberti, line 1: can we call this file `tutorial_alternative_isoforms.py`? Otherwise it might be not very clear what "isoforms" we are talking about

cancermuts/datasources.py
```diff
@@ -334,10 +335,13 @@ def cancer_studies(self):
     def add_mutations(self, sequence, metadata=[]):
         _cBioPortal_supported_metadata = ['cancer_type', 'cancer_study', 'genomic_coordinates', 'genomic_mutations']
 
+        if not sequence.is_canonical:
+            raise UnexpectedIsoformError("cBioPortal mutation annotation only supports canonical isoforms. Please use a canonical UniProt sequence")
```
mtiberti, line 339: I'd replace the last sentence with "Please use a Sequence object for a canonical isoform"

docs/tutorial.md
```diff
@@ -109,6 +111,24 @@ $ python tutorial_pancancer.py
...
+
+Cancermuts allows requesting alternative isoforms, if available in UniProt.
+To request a specific isoform, provide a isoform ID, in the form of a UniProt isoform
+identifier, to the isoform argument, as in the following example. If this is not done,
+the canonical UniProt isoform will be used. Notice that, currently, only some
+data sources support alternative isoform; those that do not support them will
+raise exceptions if a Sequence object containing an alternative isoform is provided to them.
+In this case we provide an example which use a non-canonical isoform as input in the 
+`tutorial_alternative_isoforms.py` file. It's run in the same way as with the other 
+tutorial scripts:
```
mtiberti, line 124: tutorial script (singular)

mtiberti (changes requested):

PR #226, opened 2025-07-28
Adding safeguard for non-canonical isoforms in MyVariant class and creating new local REVEL class

pablosanchezizquierdo:
fix subissue #212 from issue #43 . In this issue we will add a safeguard for non-canonical isoforms in MyVariant class and we will create a new REVEL class that will suport non-canonical isoforms and use a local REVEL database as source


cancermuts/datasources.py
```diff
@@ -1161,6 +1165,159 @@ def _get_revel_from_gc(self, mutation, gc):
...
+        super(RevelLocal, self).__init__(name='RevelLocal', version='1.0', description=description)
+
+        self._database_dir = database_dir
+
+        if database_files is None:
+            self._database_files = {
+                'revel': os.path.join(self._database_dir, 'revel_with_transcript_ids')
+            }
+        else:
+            self._database_files = database_files
```
mtiberti, line 1181: I think we can simplify this by just having one argument which is the the path to the REVEL score file, since we will always have just one

cancermuts/datasources.py
```diff
@@ -1161,6 +1165,159 @@ def _get_revel_from_gc(self, mutation, gc):
...
+
+        self._supported_metadata = {'revel_score': self._get_revel}
+
+    def add_metadata(self, sequence, md_type=['revel_score']):
+        if not sequence.uniprot_ac:
+            self.log.error("Sequence is missing UniProt accession; skipping.")
+            return
```
mtiberti, line 1188: do we need UniProt AC for REVEL? I don't think so

cancermuts/datasources.py
```diff
@@ -1161,6 +1165,159 @@ def _get_revel_from_gc(self, mutation, gc):
...
+            try:
+                metadata_functions.append(self._supported_metadata[md])
+            except KeyError:
+                self.log.warning("RevelLocal doesn't support metadata type %s" % md)
+
+        try:
+            transcript_id = self._get_transcript_id(sequence)
+        except Exception as e:
+            self.log.error(f"Failed to retrieve transcript ID for {sequence}: {e}")
+            return
```
mtiberti, line 1206: I think this should happen in the `UniProt` object instead - as per the other identifiers and similarly to what happens for other data sources

cancermuts/datasources.py
```diff
@@ -1161,6 +1165,159 @@ def _get_revel_from_gc(self, mutation, gc):
...
+                transcript_id = xref.get('id')
+                if transcript_id:
+                    transcript_id = transcript_id.split('.')[0]
+                    print(f"[REVEL] Transcript ID for {this_upac}: {transcript_id}")
+                    self.log.info(f"Using Ensembl transcript ID: {transcript_id} for {this_upac}")
+                    return transcript_id
+
+        raise KeyError(f"No Ensembl transcript ID found in UniProt entry for {this_upac}")
+
+    def _get_revel(self, mutation, transcript_id):
```
mtiberti, line 1243: it's a bit wasteful that we parse the REVEL file for every mutation - instead we should parse it once, and then keep the mutations that we need to keep. Notice that you can change the structure of `add_metadata` somewhat to make this easier

cancermuts/datasources.py
```diff
@@ -1161,6 +1165,159 @@ def _get_revel_from_gc(self, mutation, gc):
...
+                            file_aaref == aaref and
+                            file_aaalt == aaalt
+                        ):
+                            try:
+                                score = float(revel_score)
+                                mutation.metadata['revel_score'].append(DbnsfpRevel(source=self, score=score))
+                                self.log.info(f"[REVEL] Match found in line {line_num}: REVEL score = {score}")
+                            except Exception as e:
+                                self.log.warning(f"[REVEL] Could not parse REVEL score on line {line_num}: {e}")
+                            break
```
mtiberti, line 1315: I think we need to restructure this a bit 

  - as per my previous comment, I think you want to parse the REVEL file once and end up with a `pandas.DataFrame` to filter it / find your mutations
  - I tried parsing just with `pd.read_csv`, it takes ~75-80 seconds in my test. It does take a bit of memory but I think we can live with it. 
  - you shouldn't be matching on amino acid change, but on genomic mutation, since that is what we are going to score, especially for these cases, so feel free to ignore cases that don't have a genomic mutation
  - notice that you might have more than one genomic mutation per amino acid mutation (because of genetic code degeneracy) meaning that you might potentially have multiple revel scores per mutation

cancermuts/datasources.py
```diff
@@ -1161,6 +1193,107 @@ def _get_revel_from_gc(self, mutation, gc):
...
+                    if transcript_id not in row["Ensembl_transcriptid"].split(";"):
+                        continue
+
+                    score_str = row["REVEL"]
+                    if score_str in [".", "NA"]:
+                        continue
+
+                    try:
+                        score = float(score_str)
+                        mutation.metadata['revel_score'].append(DbnsfpRevel(source=self, score=score))
```
mtiberti, line 1290: we should change the name of the `DbnsfRevel` class to `Revel`, because in this case it's not from DBNSF

mtiberti (commented):
can you also check that the output is correct in the other tutorials? E.g. since you changed the name of the Revel metadata class

docs/tutorial_alternative_isoforms.py
```diff
@@ -17,6 +19,34 @@
...
+
+# annotate with REVEL using local database
+rl = RevelDatabase("/data/databases/REVEL/revel_with_transcript_ids")
+rl.add_metadata(seq)
+
+# print annotated mutation and REVEL score
+mut = seq.positions[site_seq_idx].mutations[0]
+print("Mutation:", mut)
+print("REVEL score:", mut.metadata.get('revel_score', []))
+
```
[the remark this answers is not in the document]
pablosanchezizquierdo: yes, once we have cosmic we will not need anymore the manual annotation

cancermuts/datasources.py
```diff
@@ -1222,7 +1222,7 @@ def _filter_revel_by_chromosomes(self, chrom_set):
             if missing:
                 raise ValueError(f"[REVEL] Missing columns in file: {missing}")
 
-            prefixes = tuple(f"{c}," for c in chrom_set)
+            prefixes = tuple(f"{str(c)}," for c in chrom_set)
```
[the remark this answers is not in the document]
pablosanchezizquierdo: but I need the coma if not when it's just 1 it will also match 11 so what I search is for 11, and 1,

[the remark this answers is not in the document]
pablosanchezizquierdo: so I guess I will leave it as it was at the beggining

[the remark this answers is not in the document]
mtiberti: that is my bad - I hadn't seen the comma!

yes then I agree that the `f"{c},"` works better

Issue #210, opened 2025-05-20
add support for different isoforms in `COSMIC` class


mtiberti:

mtiberti:
done in #228

PR #228, opened 2025-09-25
Filtering COSMIC by transcriptID so we fix issue + add isoform suport

### LINKED ISSUE #227: `COSMIC` class relies on COSMIC's ENSEMBL canonical status
in Cancermuts, we consider as canonical those isoforms that are considered canonical in UniProt.

However, COSMIC uses ENSEMBL for the same purpose, which might lead to a misalignment in what sequence/transcript is being used as reference between the two

pablosanchezizquierdo:
Fix #227 and #43 subissue #210


cancermuts/datasources.py
```diff
@@ -682,7 +678,10 @@ def _load_db_files(self, targeted_db_file, screenmut_db_file):
         for fi, file in enumerate(targeted_screenmut_db_files):
             self.log.info(f"Parsing database file {targeted_screenmut_db_filenames[fi]}...")
             try:
-                targeted_screenmut_dataframes.append(pd.read_csv(file, sep='\t', dtype='str', na_values='NS', usecols=self._use_cols_database_files, encoding=self._encoding))
+                df_tmp = pd.read_csv(file, sep='\\t', dtype='str', na_values='NS', usecols=self._use_cols_database_files, encoding=self._encoding)
+                if 'TRANSCRIPT_ACCESSION' in df_tmp.columns:
```
mtiberti, line 682: have you found cases in which `TRANSCRIPT_ACCESSION` is not availalbe or is it more of a just in case?

and in case how should we handle it?

tbh, if it's a more of a "just in case", we should just allow it to raise an exception that is then handled with the except below

cancermuts/datasources.py
```diff
@@ -729,7 +713,8 @@ def _parse_db_files(self, gene_id, genome_assembly_version = 'GRCh38',
                        cancer_site_subtype_1=None,
                        cancer_site_subtype_2=None,
                        cancer_site_subtype_3=None,
-                       metadata=[]):
+                       metadata=[],
+                       transcript_accession=None):
```
mtiberti, line 717: can this argument come before the metadata section in the list of arguments? e.g.
```
self, gene_id, genome_assembly_version = 'GRCh38', transcript_accession=None, ...)
```

mtiberti: also, should we consider making this a mandatory argument?

At this point I don't think it makes much sense to filter if we don't have it

cancermuts/datasources.py
```diff
@@ -773,7 +759,14 @@ def _parse_db_files(self, gene_id, genome_assembly_version = 'GRCh38',
 
             df = self._load_db_files(filtered_targeted_database_file, filtered_screenmut_database_file)
 
-            df = df[ (df['IS_CANONICAL'] == 'y') ]
+        if transcript_accession:
```
mtiberti, line 762: probably needs to be removed if it becomes a mandatory arg

cancermuts/datasources.py
```diff
@@ -773,7 +759,14 @@ def _parse_db_files(self, gene_id, genome_assembly_version = 'GRCh38',
...
+            self.log.info(f"Filtering by transcript_accession={transcript_accession}")
+            if 'TRANSCRIPT_ACCESSION' in df.columns:
```
mtiberti, line 764: same question as before - are there cases in which this is not the case?

cancermuts/datasources.py
```diff
@@ -773,7 +759,14 @@ def _parse_db_files(self, gene_id, genome_assembly_version = 'GRCh38',
...
+                df = df[df['TRANSCRIPT_ACCESSION'] == transcript_accession]
+                if df.empty:
+                    self.log.warning(f"No COSMIC rows for gene {gene_id} with TRANSCRIPT_ACCESSION={transcript_accession}; falling back to gene-level rows")
```
mtiberti, line 767: can we change the wording from rows to mutations? So it's more clear for users

also I don't think we should fall-back, this is error prone; I think we should just continue and return an empty dataframe

cancermuts/datasources.py
```diff
@@ -870,6 +859,12 @@ def add_mutations(self, sequence, genome_assembly_version='GRCh38',
         else:
             gene_id = sequence.gene_id
 
+        transcript_accession = sequence.aliases.get('ensembl_transcript_id')
+        if transcript_accession:
+            self.log.info(f"Using Ensembl transcript for COSMIC filter: {transcript_accession}")
+        else:
+            self.log.info("No Ensembl transcript found in sequence aliases; using gene-level COSMIC rows")
```
mtiberti, line 866: again, we should avoid using this fallback strategy since it's error prone

docs/tutorial_alternative_isoforms.py
```diff
@@ -19,16 +19,40 @@
...
-                "manual;p.Val820Met;mutation;;;"
-                "hg19,11:g.46456582C>T\n")
-
-# add mutation with ManualAnnotation class
-ma = ManualAnnotation(csv)
-ma.add_mutations(seq, metadata=['genomic_mutations'])
+print("Transcript accession for filtering (if any):",
+      seq.aliases.get('transcript_accession') or seq.aliases.get('ensembl_transcript_id'))
+
+# COSMIC now suports non-canonical alternative isofrom
```
mtiberti, line 25: you can skip this comment

docs/tutorial_alternative_isoforms.py
```diff
@@ -50,15 +74,8 @@
 cosmic = COSMIC(targeted_database_file='/data/databases/cosmic-v102/Cosmic_CompleteTargetedScreensMutant_v102_GRCh38.tsv',
                 screen_mutant_database_file='/data/databases/cosmic-v102/Cosmic_GenomeScreensMutant_v102_GRCh38.tsv',
                 classification_database_file='/data/databases/cosmic-v102/Cosmic_Classification_v102_GRCh38.tsv',
-                transcript_database_file='/data/databases/cosmic-v102/Cosmic_Transcripts_v102_GRCh38.tsv',
                 database_encoding='latin1', lazy_load_db=True)
```
mtiberti, line 77: I'm not sure what is the point of keeping this given the previous code block we are adding

docs/tutorial.md
```diff
@@ -849,7 +849,8 @@ the Table module:
 ```
 
 We can then manipulate the dataframe as we see fit, e.g. by saving it
-as a csv file:
+as a csv file (note that it can be different from the precalculated ones in 
+the tutorial due to database updates):
```
mtiberti, line 853: this is fine, but you also need to work on the COSMIC section of this file to update the function calls and check if we say re supporting alternative isoforms

mtiberti (changes requested):

cancermuts/datasources.py
pablosanchezizquierdo: there should not be any case

mtiberti (changes requested):

cancermuts/datasources.py
```diff
@@ -773,8 +757,13 @@ def _parse_db_files(self, gene_id, genome_assembly_version = 'GRCh38',
 
             df = self._load_db_files(filtered_targeted_database_file, filtered_screenmut_database_file)
 
-            df = df[ (df['IS_CANONICAL'] == 'y') ]
 
+        self.log.info(f"Filtering by transcript_accession={transcript_accession}")
+        df = df[df['TRANSCRIPT_ACCESSION'] == transcript_accession]
+        if df.empty:
+            self.log.warning(f"No COSMIC mutations for gene {gene_id} with TRANSCRIPT_ACCESSION={transcript_accession}; falling back to gene-level rows")
```
mtiberti, line 764: fix the warning message accordingly to how it works now

PR #232, opened 2025-10-21
Added condition in GenomicMutation to remove definition nans values

AlessiaCampo:
Found a solution to not parse `GenomicMutation` in the metadata if the `definition` of the genomic mutation is `nan`. After careful consideration of how the classes in `metadata.py` are structured- the solution exploits the `__eq__()` method to check if the `definition` value of the genomic mutation is `nan`. This should be a consistent with the other classes where `nan` values for other metadata type are removed.

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -1660,7 +1660,7 @@ class COSMIC(DynamicSource, object):
         if df.empty:
             self.log.warning(f"No COSMIC mutations for gene {gene_id} with TRANSCRIPT_ACCESSION={transcript_accession}; returning empty results")
             return [], out_metadata
-            
+
         if cancer_types is not None:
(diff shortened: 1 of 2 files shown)


mtiberti:
that would be a way to solve the problem. Alternatively,

I was looking into the `COSMIC` class, we have

```
            if do_genomic_mutations:
                if gd is None:
                    self.log.warning("couldn't annotate genomic mutation")
                    gm = None
                else:
                    gm = [gd[0], r['HGVSG']]
                out_metadata['genomic_mutations'].append(gm)
```

another way would be to check if the genomic mutation is there using `pd.isna` instead of `is` - that would make it `None` in this code.

I think this would be an even better solution as it relies on the standard mechanism we have for the other metadata for which `None` is used to signal "no metadata" instead of any null value.

Makes sense?

ps: I would also check what is happening with the genomic coodinates metadata - my initial exploration suggested there was some funny businness going on for similar cases

PR #233, opened 2025-10-29
Add ptm_sources column to final metatable (#230)

### LINKED ISSUE #230: Add column about PTM sources in final metatable

drgbea:
fixes #230 

table.py has been modified to include ptm_sources in the metatable


cancermuts/table.py
```diff
@@ -165,7 +165,8 @@ def __init__(self, labels=None, ptm_colors=None, y_ptm=1.02):
         headers['position']    = SequencePosition.header
         headers['wt']          = 'ref_aa'
         headers['mutated']     = 'alt_aa'
-        headers['mut_sources'] = 'sources'
+        headers['mut_sources'] = 'mut_sources'
```
mtiberti: can we keep this `sources` as is?

just because it potentially breaks compatibility with tools that use the metatable

cancermuts/table.py
```diff
@@ -189,7 +190,7 @@ def to_dataframe(self, sequence, mutation_metadata=["cancer_study", "cancer_type
 
         header += [ self.headers[p] for p in sequence_properties ]
         sequence_properties_col = list(range(sequence_properties_cols_start, len(header)))
-        header += [self.headers['wt'], self.headers['mutated'], self.headers['mut_sources']]
+        header += [self.headers['wt'], self.headers['mutated'], self.headers['mut_sources'], self.headers['ptm_sources']]
```
mtiberti: this would be better positioned after the position properties (=ptm columns)

mtiberti (changes requested):
can you also rerun all the tutorial scripts as they are and update their output?

thanks!

Issue #244, opened 2025-11-19
Cancermuts only retrieves germline classification for clinvar variants


elenikiachaki:
As of now, we only include germline classification data (classification, review status, condition) for the clinvar variants in the metatable.
We should expand the metatable with new columns that will cover the **somatic classification** and **clinical impact classification**, if provided

Example of variant with 2 out of the 3 classification types:

<img width="1900" height="495" alt="Image" src="https://github.com/user-attachments/assets/dfd9595c-1101-4b79-8deb-debb6f04c34c" />

mtiberti:
done in #245

PR #245, opened 2025-11-21
Added new metatable columns for somatic classification data

elenikiachaki:
PR that addresses issue #244 
With the following changes, apart from **Germline classification** information, we additionally annotate:
- Oncogenicity information
- Clinical impact information
for clinvar variants.

A good example for testing (few variants, among which one with oncogenicity classification: **CD70, CD70_HUMAN, NP_001243**


cancermuts/metadata.py
```diff
@@ -616,21 +634,65 @@ def get_stars(self):
         return self.stars
 
     def __repr__(self):
-        return f"<ClinVarReviewStatus {self.stars} (from: {self.status}) from {self.source.name}>"
+        return f"{self.header} {self.stars} (from: {self.status}) from {self.source.name}>"
```
mtiberti: can we use a slightly different `__repr__`? It should look something like

```
ClinvarReviewStatus(x=..., y=...)
```

PR #251, opened 2026-02-26
Add GlyGen datasource support for glycosylation site annotations (fix250_glygen_data_source)

drgbea:
Implemented a new GlyGen datasource that parses dataset GLY_000038 and adds glycosylation annotations to sequences. Subtypes are generated from glycosylation type + glycan name and are now correctly exported in the metatable.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -691,10 +691,10 @@ GlyGen integrates glycoproteomics data from multiple experimental studies and pr
 
 To use GlyGen, a local copy of the dataset must be available (see the installation guide for details). In this tutorial we use a filtered version of the GlyGen dataset that contains only experimentally validated glycosylation sites.
 
-We first create the GlyGen data source object by specifying the location of the dataset file:
+We first create the GlyGen data source object by specifying the location of the dataset file, and its filename, as we will be using a filtered version of the original database:
 
 ```py
->>> gg = GlyGen('/data/databases/GlyGen/')
+>>> gg = GlyGen('/data/databases/GlyGen/', database_file='human_proteoform_glycosylation_sites_uniprotkb_filtered.csv')
 ```
 Once the object is created we can add the position properties to our sequence object.
 
--- a/docs/tutorial.py
+++ b/docs/tutorial.py
@@ -110,8 +110,8 @@ print(seq.positions[4].properties)
 print(seq.positions[28].properties)
 
 # add annotations from GlyGen
-gg = GlyGen('/data/databases/GlyGen/')
-gg.add_position_properties(seq)
+gg = GlyGen('/data/databases/GlyGen/', database_file='human_proteoform_glycosylation_sites_uniprotkb_filtered.csv')
+gg.add_position_properties(seq, )
 
 # add annotations from MobiDB
 mdb = MobiDB()
--- a/docs/tutorial_alternative_isoforms.py
+++ b/docs/tutorial_alternative_isoforms.py
@@ -109,7 +109,7 @@ except UnexpectedIsoformError:
     print("PhosphoSite annotations will not be added, as a non-canonical isoform has been provided")
 
 # GlyGen does not support non-canonical isoforms
-gg = GlyGen('/data/databases/GlyGen/')
+gg = GlyGen('/data/databases/GlyGen/', database_file='human_proteoform_glycosylation_sites_uniprotkb_filtered.csv')
 
 try:
     gg.add_position_properties(seq)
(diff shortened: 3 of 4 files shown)


cancermuts/datasources.py
```diff
@@ -1938,6 +1939,68 @@ def add_position_properties(self, sequence, properties=None):
...
+    @logger_init
+    def __init__(self, database_dir, database_files=None):
+        description = "GlyGen Dataset 000038"
+        super(GlyGen, self).__init__(name='GlyGen', version='2.10.1', description=description)
+        
+        self.database_dir = database_dir
+        self.database_files = database_files
+
+    def _parse_db_file(self, protein_id):
+        filepath = os.path.join(self.database_dir, "human_proteoform_glycosylation_sites_uniprotkb.csv")
```
mtiberti: the filename should not be hard-coded, we store it in `self.database_files` - let's use that

cancermuts/datasources.py
```diff
@@ -1938,6 +1939,68 @@ def add_position_properties(self, sequence, properties=None):
...
+            uniprot_ac = row['src_xref_id']
+            record['uniprot_ac'] = uniprot_ac
+            record['position'] = int(glycosylation_site)
+            record['glycosylation_type'] = glycosylation_type
+            record['carb_name'] = carb_name
+            records.append(record)
+
+        return records
+    
+    def add_position_properties(self, sequence):
```
mtiberti: first of all, we need to add a check to see if the sequence is isoform canonical and exit if not. See `cancermuts.datasources.MobiDB.add_position_properties` - we should do so in the same way

cancermuts/datasources.py
```diff
@@ -1938,6 +1939,68 @@ def add_position_properties(self, sequence, properties=None):
...
+        filepath = os.path.join(self.database_dir, "human_proteoform_glycosylation_sites_uniprotkb.csv")
+        records = []
+        df = pd.read_csv(filepath)
+        protein_id = protein_id.split("-")[0]
+        protein_df = df[df['src_xref_id'] == protein_id]
+        for _, row in protein_df.iterrows():
+            record = {}
+            glycosylation_site = row["glycosylation_site_uniprotkb"]
+            glycosylation_type = row["glycosylation_type"]
+            glycosylation_type = glycosylation_type.rstrip('-linked.')
```
mtiberti: this is something you can do on a per-column basis:

```
In [2]: df = pd.DataFrame({'a': ['a.', 'bb.', 'c..']})

In [3]: df
Out[3]:
     a
0   a.
1  bb.
2  c..

In [4]: df['a'].str.rstrip('.')
Out[4]:
0     a
1    bb
2     c
Name: a, dtype: object
```

in general, in `pandas` it's much more efficient to perform vectorized operations (across columns) than iterating over rows, which is in general discouraged. See e.g. `pandas.DataFrame.apply`

cancermuts/datasources.py
```diff
@@ -1938,6 +1939,49 @@ def add_position_properties(self, sequence, properties=None):
...
+            subtype = f"{row['glycosylation_type']}-{row['carb_name'].rstrip('.')}"
+            position_obj = sequence.positions[idx]
+
+            already_annotated = False
+            for prop in position_obj.properties:
+                if isinstance(prop, GlycosylationSite):
+                    prop.sources.append(self)
+                    prop.add_subtype(subtype)
+                    already_annotated = True
+                    property_obj = prop
```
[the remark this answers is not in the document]
drgbea: I deleted `property_obj = prop`
I forgot to delete it before the last push

docs/metatable.csv
```diff
@@ -1,6 +1,6 @@
...
 0,1,M,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
-1,2,P,R,,,,,,,,,,,,,,ClinVar,,,"hg19,16:87426038-87426038, hg38,16:87392432-87392432","hg19,16:g.87426038C>G, hg38,16:g.87392432C>G",0.145,,,0.0000000000000000,,,,2358122,Uncertain significance,not specified,1,,,,,,
+1,2,P,R,,,,,,,,,,,,,,ClinVar,,,"hg19,16:87426038-87426038, hg38,16:87392432-87392432","hg19,16:g.87426038C>G, hg38,16:g.87392432C>G",0.145,,,,,,,2358122,Uncertain significance,not specified,1,,,,,,
```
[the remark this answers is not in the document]
mtiberti: got it thanks, then it's not an issue related to this PR, so we can continue. But probably deserves being investigate independently

Issue #260, opened 2026-03-17
ClinVar assembly-specific genomic annotation causes cancermuts crash for OPTN variant 703420


elenikiachaki:
**Description:**
During MAVISp automatization, the cancermuts step was crashing for OPTN while generating the pancancer metatable.

The problematic variant was:
- Gene/protein: OPTN
- ClinVar ID: 703420
- Variant: NM_021980.4(OPTN):c.964G>A (p.Glu322Lys)

**The error was:**

`AttributeError: 'GenomicMutation' object has no attribute 'coord_start'`

This happened in `datasources.py` while building `genomic_coordinates`, at: `gm.coord if gm.is_snv else gm.coord_start`

**Why it happens**:

For this ClinVar entry, the API returns different genomic HGVS annotations for the two assemblies:
- GRCh38: NC_000010.11:g.13124076G>A
- GRCh37: NC_000010.10:g.13166076=

After checking the locus, this is expected: in GRCh38 the reference base is G, so G>A is a valid SNV, whereas in GRCh37 the reference base is already A, so ClinVar reports = (no change).

In GenomicMutation in `metadata.py`, the GRCh37 annotation does not match any supported mutation format, so it falls into the final fallback else. However, in that branch not all attributes are initialized to **None**, and later `datasources.py` assumes that every non-SNV object has `coord_start`, causing the crash.

**Expected behavior:** cancermuts should not crash in this case, but rather keep the valid assembly-specific genomic annotation(s) and skip or return None for unsupported/inconsistent ones.

elenapapaleo:
this should be solved by the last PR we did t fix the irregular genomic annotations

PR #261, opened 2026-03-18
Excluded irregular genomic annotations from clinvar metadata

elenikiachaki:
PR that addresses #260 
This PR fixes a crash in ClinVar annotation caused by assembly-specific genomic entries that do not match any of the supported `GenomicMutation` patterns in `metadata.py` (SNV, inversion, insdel).

With this change, unsupported genomic annotations are skipped from the variant’s ClinVar genomic metadata, while valid annotations for the same variant are still kept.

**Example:** for ClinVar ID 703420 (NM_021980.4(OPTN):c.964G>A, p.Glu322Lys), the GRCh38 annotation `NC_000010.11:g.13124076G>A` is kept, while the GRCh37 annotation `NC_000010.10:g.13166076=` is not included in genomic_mutations / genomic_coordinates.

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -1452,17 +1452,23 @@ class ClinVar(DynamicSource, object):
                         out_metadata[rs_key].append(rs_dict[xml_key])
                     else:
                         out_metadata[rs_key].append(None)
-
+            valid_gms = []
+            for gm in data["genomic_annotations_obj"].values():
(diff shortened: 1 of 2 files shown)


cancermuts/datasources.py
pablosanchezizquierdo, line 1: This is fixing the issue, now we avoid malformed clinvar entries

pablosanchezizquierdo (approved):

Issue #264, opened 2026-03-30
Change `Mutation` class into `ProteinVariant` class to support different types of variants


mtiberti:

mtiberti:
done in #265

PR #265, opened 2026-04-07
Creation of ProteinVariant class and related changes

elenikiachaki:
Addresses issue #264

**core.py:**
   - added `ProteinVariant` class
   - moved variant from `SequencePosition` to `Sequence`
   - added `Sequence.add_variant()`

**datasources.py:**
   - updated datasources code to create `ProteinVariant` obj and store in `Sequence` obj

### DIFF
--- a/cancermuts/core.py
+++ b/cancermuts/core.py
@@ -21,7 +21,7 @@
 Core cancermuts classes --- :mod:`cancermuts.core`
 ================================================================
 core classes that define the basic framework to handle a sequence
-and its mutations.
+and its variants.
 
 """
 
@@ -34,8 +34,7 @@ from .log import logger_init
 class Sequence(object):
     """The most fundamental class of Cancermuts, this class starts from a
     protein sequence definition. It acts as a collection of ordered 
-    `SequencePosition` objects which depend on the specific
-    sequence itself.
+    `SequencePosition` objects and protein variant objects.
 
     Parameters
     ----------
@@ -90,6 +89,7 @@ class Sequence(object):
         self.aliases["uniprot_acc"] = self.uniprot_ac
         self.source = source
         self.sequence = sequence
+        self.variants = []
         self.positions = []
         self.sequence_numbering = []
         self.properties = {}
@@ -112,7 +112,6 @@ class Sequence(object):
         i:obj:`int`
             sequence number corresponding to the idx-th position
         """
-
 
     def seq2index(self, seqn):
         """
@@ -159,12 +158,39 @@ class Sequence(object):
             add_type = "new category"
             self.log.debug("adding property %s to sequence of %s (%s)" % (str(prop), self.gene_id, add_type))
 
+    def add_variant(self, var):
+        """
+        Adds variant to a :obj:`Sequence` object. If the variant is already present, 
+        source and metadata will be added to the already-present variant. Otherwise
+        the variant is added as new.
+
+        Parameters
+        ----------
+        var : :obj:`cancermuts.core.ProteinVariant`
+            new ProteinVariant object to be added to the Sequence
+        """
+
+        if var not in self.variants:
+
+            self.variants.append(var)
+            self.log.info("Adding variant %s to sequence %s" % (str(var), self.__repr__()))
+        else:
+            self.log.info("Variant %s already in sequence %s; will just add sources and metadata" % (str(var), self.__repr__()))
+            idx = self.variants.index(var)
+            self.variants[idx].sources.extend(var.sources)
+            for k in var.metadata:
+                if k in self.variants[idx].metadata:
+                    self.variants[idx].metadata[k].extend(var.metadata[k])
+                    self.log.debug("    metadata %s was extended" % k)
+                else:
+                    self.variants[idx].metadata[k] = var.metadata[k]
+                    self.log.debug("    metadata %s was added anew" % k)
+
 
 class SequencePosition(object):
-    """This class describe a certain sequence position in a protein.
-    SequencePositions usually belong to a Sequence object.
-    It is possible to annotate a Sequence position with either a Mutation
-    or a PositionProperty.
+    """This class describes a certain sequence position in a protein.
+    SequencePositions belong to a Sequence object and represent
+    the wild-type residue at a given sequence coordinate.
 
     Attributes
     ----------
@@ -173,16 +199,11 @@ class SequencePosition(object):
         single-letter code wild-type residue for this position
     sequence_position : :obj:`int`
         number corresponding to the sequence position for this position
-    mutations : :obj:`list` of `cancermuts.core.Mutation` objects
-        list of mutations for this sequence position (if any)
-    properties : :obj:`dict`
-        dictionary encoding position properties. This dictionary needs to have
-        :obj:`str` as key and `cancermuts.properties.PositionProperty` as value.
     """
     description = 'Position'
     header = "aa_position"
     @logger_init
-    def __init__(self, wt_residue_type, sequence_position, mutations=None, properties=None):
+    def __init__(self, wt_residue_type, sequence_position):
         """Constructor for the SequencePosition class.
 
         Parameters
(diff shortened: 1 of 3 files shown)


cancermuts/core.py
```diff
@@ -90,6 +89,7 @@ def __init__(self, gene_id, uniprot_ac, sequence, source, isoform=None, is_canon
         self.aliases["uniprot_acc"] = self.uniprot_ac
         self.source = source
         self.sequence = sequence
+        self.variants = []
         self.positions = []
```
mtiberti: this probably needs to be removed when we will have the `VariantRegister`

cancermuts/core.py
```diff
@@ -90,6 +89,7 @@ def __init__(self, gene_id, uniprot_ac, sequence, source, isoform=None, is_canon
         self.aliases["uniprot_acc"] = self.uniprot_ac
         self.source = source
         self.sequence = sequence
+        self.variants = []
```
mtiberti: this will need to be `VariantRegister`, eventually

cancermuts/core.py
```diff
@@ -159,12 +158,39 @@ def add_property(self, prop):
...
+
+        Parameters
+        ----------
+        var : :obj:`cancermuts.core.ProteinVariant`
+            new ProteinVariant object to be added to the Sequence
+        """
+
+        if var not in self.variants:
+
+            self.variants.append(var)
```
mtiberti: this will change with the `VariantRegister`

cancermuts/core.py
```diff
@@ -191,119 +212,69 @@ def __init__(self, wt_residue_type, sequence_position, mutations=None, propertie
...
+            sources the variant was derived from
+        metadata : :obj:`dict`
+            Dictionary encoding variant-associated metadata.
     """
 
     @logger_init
-    def __init__(self, sequence_position, mutated_residue_type, sources=None, metadata=None):
-        """Constructor for the Mutation class.
-
+    def __init__(self, sequence, start, end, ref, alt, variant_type, sources=None, metadata=None):
```
mtiberti: do we really need a `Sequence`?

mostly I'm asking because references become circular: sequence -> variants -> variant -> sequence -> ...

cancermuts/core.py
mtiberti: we should also check that the position and residue types correspond to the wt - and otherwise raise an exception

cancermuts/core.py
```diff
@@ -159,12 +158,39 @@ def add_property(self, prop):
...
+            for k in var.metadata:
+                if k in self.variants[idx].metadata:
+                    self.variants[idx].metadata[k].extend(var.metadata[k])
+                    self.log.debug("    metadata %s was extended" % k)
+                else:
+                    self.variants[idx].metadata[k] = var.metadata[k]
+                    self.log.debug("    metadata %s was added anew" % k)
+
 
 class SequencePosition(object):
```
mtiberti: can you just remove this class? I don't think we'll need it for how things are now

cancermuts/core.py
```diff
@@ -191,119 +212,69 @@ def __init__(self, wt_residue_type, sequence_position, mutations=None, propertie
...
+            1-based start coordinate on the protein sequence
+        end : :obj:`int`
+            1-based end coordinate on the protein sequence
+        ref : :obj:`str`
+            reference amino-acid sequence
+        alt : :obj:`str`
+            altered amino-acid sequence
+        variant_type: :obj:`str`
+            Type of protein variant. Supported values currently include
+            ``"substitution"``, ``"deletion"``, ``"insertion"``, and ``"delins"``.
```
mtiberti: I think this should not be determined as an argument - it should just be derived by the constructor from the other parameters, so we're guaranteed it's always consistent

cancermuts/core.py
```diff
@@ -191,119 +212,69 @@ def __init__(self, wt_residue_type, sequence_position, mutations=None, propertie
...
 
-        self.sequence_position = sequence_position
-        self.mutated_residue_type = mutated_residue_type
+        """
+        self.sequence = sequence
+        self.start = start
+        self.end = end
+        self.ref = ref
+        self.alt = alt
+        self.variant_type = variant_type
```
mtiberti: see previous comment

cancermuts/core.py
```diff
@@ -313,14 +284,29 @@ def __init__(self, sequence_position, mutated_residue_type, sources=None, metada
...
+        elif self.variant_type == "deletion":
+            if self.start == self.end:
+                return "%s%ddel" % (self.ref, self.start)
+            return "%s%d_%ddel" % (self.ref, self.start, self.end)
+        elif self.variant_type == "insertion":
+            return "%d_%dins%s" % (self.start, self.end, self.alt)
+        elif self.variant_type == "delins":
+            return "%s%d_%ddelins%s" % (self.ref, self.start, self.end, self.alt)
+        else:
+            raise TypeError("Unsupported variant type %s" % self.variant_type)
```
mtiberti: I think we should catch this at constructor level and raise there

cancermuts/core.py
```diff
@@ -313,14 +284,29 @@ def __init__(self, sequence_position, mutated_residue_type, sources=None, metada
...
+                                    ", ".join([s.name for s in self.sources]))         
+    def __eq__(self, other):
+        if not isinstance(other, ProteinVariant):
+            return False
+        return (self.sequence == other.sequence and
+                self.start == other.start and
+                self.end == other.end and
+                self.ref == other.ref and
+                self.alt == other.alt and
+                self.variant_type == other.variant_type)
```
mtiberti: I don't think `variant_type` should be part of this comparison - since it's supposed to depend on the rest of the data

cancermuts/core.py
mtiberti: we could make this a `property` (see e.g. `@property` decorator) with a getter function that acts dynamically. We could also cache this, it's a bit more cumbersome and I don't think it would really be necessary

cancermuts/datasources.py
```diff
@@ -2123,10 +2123,9 @@ def add_metadata(self, sequence, md_type=['revel_score']):
             except KeyError:
                 self.log.warning("MyVariant doesn't support metadata type %s" % md_type)
 
-        for pos in sequence.positions:
-            for mut in pos.mutations:
-                for add_this_metadata in metadata_functions:
-                    add_this_metadata(mut)
+        for var in sequence.variants:
```
mtiberti, line 2125: since this is necessary, we probably also need to add some iterator behavior on the `VariantRegister`, and so define `__iter__` and `__next__`. That means we need some strategy to order the variants to make it predictable (as we do already anyway in `VariantRegister`)

mtiberti (changes requested):

cancermuts/core.py
```diff
@@ -87,15 +89,11 @@ def __init__(self, gene_id, uniprot_ac, sequence, source, isoform=None, is_canon
...
-        self.variants = []
-        self.positions = []
-        self.sequence_numbering = []
+        self.sequence_numbering = list(range(1, len(self.sequence) + 1))
         self.properties = {}
-        for i,p in enumerate(self.sequence):
-            self.positions.append(SequencePosition(p, i+1))
-            self.sequence_numbering.append(i+1)
 
     def index2seq(self, idx):
```
elenikiachaki: Why do we have this empty method?

cancermuts/core.py
```diff
@@ -284,29 +256,178 @@ def __init__(self, sequence, start, end, ref, alt, variant_type, sources=None, m
...
+        self._register[hgvs] = variant
+
+    def _is_valid_variant_string(self, item):
+        return (re.match(self._substitution_format, item)
+                or re.match(self._deletion_format, item)
+                or re.match(self._insertion_format, item)
+                or re.match(self._duplication_format, item)
+                or re.match(self._delins_format, item))
+
+    def __getitem__(self, item):
```
elenikiachaki, line 345: It only supports one-letter notations like: `H39F`, `H39_K40insGG`, `H39_K40dup`, `H39_K40delinsGG`. 
I don't know if it should be able to accept and convert from three-letter to one-letter notations or not. As far as I understand, for substitutons this conversion is handled in datasources.py, so by the time variants are passed into `core.py` they are already represented in one-letter format.

cancermuts/core.py
mtiberti: it's a mistake. I think I meant to add it for the sake of completeness but then never did, and it was never necessary. I think it can be removed unless we have a need for it now (which I doubt at the moment)

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
+        end : :obj:`int`
+            1-based end coordinate on the protein sequence
+        ref : :obj:`str`
+            reference amino-acid sequence
+        alt : :obj:`str`
+            altered amino-acid sequence
+        start_res : :obj:`str`
+            Wild-type amino-acid at the start coordinate
+        end_res : :obj:`str`
+            Wild-type amino-acid at the end coordinate
```
mtiberti: do we really need these arguments - can't they be derived from `ref`?

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
+            Dictionary encoding variant-associated metadata.
 
-        if mut not in self.mutations:
+        """
+        self.start = start
+        self.end = end
+        self.ref = ref
+        self.alt = alt
+        self.start_res = start_res
+        self.end_res = end_res
```
mtiberti: as per previous comment - if they can't be derived from `ref` for some reason I can't see now, we should check at least if they are consistent with `ref` and `alt`

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
+    def variant_type(self):
+        if (self.start == self.end and len(self.ref) == len(self.alt) == 1
+            and self.ref != self.alt):
+            return "substitution"
+        elif self.ref != "" and self.alt == "":
+            return "deletion"
+        elif self.ref == "" and self.alt != "":
+            return "insertion"
+        elif self.ref != "" and self.alt == self.ref:
+            return "duplication"
```
mtiberti: this is a bit confusing to me - looking at this condition is, it looks like a silent substitution of multiple residues (since ref=alt) rather than a duplication (in which the same stretch of residues is repeated more than once)

do we have reasons to support duplications explicitly and treat them differently than regular insertions? In the end they are a special type of insertion

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
-            self.mutations.append(mut)
-            #print self.mutations[-1].metadata
-            self.log.info("Adding mutation %s to position %s" % (str(mut), self.__repr__()))
+        self.variant_type
+
+    @property
+    def variant_type(self):
+        if (self.start == self.end and len(self.ref) == len(self.alt) == 1
+            and self.ref != self.alt):
+            return "substitution"
```
mtiberti: can we call this missense?

an indel can also be a substitution

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
+                             f"position {var.start}, got {self.sequence[var.start - 1]}")
 
-class SequencePosition(object):
-    """This class describe a certain sequence position in a protein.
-    SequencePositions usually belong to a Sequence object.
-    It is possible to annotate a Sequence position with either a Mutation
-    or a PositionProperty.
+        if self.sequence[var.end - 1] != var.end_res:
+            raise ValueError(f"WT mismatch for {var.hgvs}: expected {var.end_res} at "
+                             f"position {var.end}, got {self.sequence[var.end - 1]}")
```
mtiberti: Here I'm not sure why we check just start and res

we have all the positions, ref, and the sequence, can't we just check something akin to `self.sequence[var.start:var.stop+1] == var.ref`? (unless it's an insertion ofc - in that case we just need to check if the position falls within the sequence)

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
-                    self.log.debug("    metadata %s was added anew" % k)
+            raise ValueError(f"Cannot determine variant type from start={self.start}, "
+                             f"end={self.end}, ref='{self.ref}', alt='{self.alt}'")
 
-    def add_property(self, prop):
+    @property
+    def hgvs(self):
+        return f"p.{self}"
+
+    def __str__(self):
```
mtiberti: I would swap these two definitions

have a `hgvs` that properly builds the `hgvs` string

and a `__str__` that just returns `hgvs`

in this way `hgvs` can be a `@property` and if we need down the line we can easily change `__str__`

cancermuts/core.py
mtiberti: one-letter format is fine, in my head it was even just supposed to support missense mutations (but it's also good if it supports more)

it should also support HGVS though

at least in my head, items in the register are registered by hgvs - so it should be trivial to get variants specifying hgvs. If we get a non-HGVS-conforming string, then we should assume it's a single-letter notation (MAVISp notation, e.g. `H39P`) and convert it to HGVS, and use that to pull information out of the register

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
 
-        if prop.category in self.properties:
-            self.log.info("property %s was replaced with %s" % (self.properties[prop.category], prop))
-        else:
-            self.log.info("added property %s" % str(prop))
+        Parameters
+        ----------
+        item : :obj:`str`
+            Variant string in one-letter format (e.g. ``H39F``) or HGVS format
+            (e.g. ``p.H39F``).
```
mtiberti: notice that `p.H39F` is not HGVS-compliant (as I think you know) - we should really use HGVS when we receive HGVS

cancermuts/core.py
mtiberti: ps: there also is a `hgvs` Python package which could be helpful - I tried using having a quick look at it before, I couldn't figure it out quickly but we could be able to get something from there

cancermuts/core.py
```diff
@@ -159,168 +156,278 @@ def add_property(self, prop):
...
+        if item.startswith("p."):
+            bare_item = item[2:]
+            if self._is_valid_variant_string(bare_item):
+                item_hgvs = item
+            else:
+                raise KeyError(f"Unsupported variant format: {item}")
+        elif self._is_valid_variant_string(item):
+            item_hgvs = "p." + item
+        else:
+            raise KeyError(f"Unsupported variant format: {item}")
```
mtiberti: I think this could be something like

```
if self._is_valid_hgvs(item):
    return self._register[item]
elif self._is_valid_mavisp_fmt(item):
    return self._register[self._mavisp2hgvs(item)]
```

mtiberti (changes requested):

cancermuts/core.py
elenikiachaki: `HGVS` treats duplication as a sepearate category ([HGVS nomenclature](https://hgvs-nomenclature.org/stable/recommendations/protein/duplication/)). Therefore duplications will be parsed in `datasources.py` from COSMIC, CBioportal and ClinVar by matching a pattern like `12Ser_15Aladup`. 

But I agree that they are indeed a special kind of insertion and the code would be simplified if we stored them in cancermuts as insertions. Should I therefore remove the duplication `variant_type`?

cancermuts/core.py
elenikiachaki: My confusion was exactly here: how do we know `ref `without reconstructing it from the wt sequence?

For several HGVS classes, the full reference stretch is not present in the mutation string itself:

- multi-residue deletions: `p.Ser14_Ala19del`
- multi-residue delins: `p.Ser123_Ala130delinsAla`
- multi-residue duplications: `p.Ser14_Ala19dup`

In these cases, HGVS gives only the boundary residues/positions, not the full affected peptide, so `ref `cannot be derived from the HGVS string alone.

For insertions, according to how I have defned these arguments, `ref == ""` and `alt == inserted sequence`, so there is no reference segment to compare against sequence. In that case, the only sanity check I could think of was to verify the flanking WT residues (`start_res` / `end_res`) against the sequence.

That is why I introduced start_res and end_res.

cancermuts/core.py
elenikiachaki: but do we want to use p.Trp24Cys syntax or NP_003997.1:p.Trp24Ter  as HGVS compliant? I think the latter one is the correct but the first is shorter to use as key in the register

cancermuts/core.py
mtiberti: yes - I think it makes sense to treat them as inertions internally, we can always use the connected hgvs to track back if they were originally found as duplications

cancermuts/core.py
mtiberti: I see what you mean - I thought we could use the original sequence to pull the residues that are missing and validate the `ref`, is there something preventing us to? I think it would be more robust and intentional than having start_res/end_res as argument

got it re insertions, I think it's a good idea, but again I don't think it requires having the start_res and end_res arguments

cancermuts/core.py
mtiberti: I think I would use the first, since the `VariantRegister` is still always anchored to a `Sequence` that we know already, unless there are strong reasons otherwise (i.e. you want to use the `hgvs` package which requires a reference sequence for what I can tell)

cancermuts/core.py
```diff
@@ -159,168 +142,276 @@ def add_property(self, prop):
...
+        self.variant_type
+
+    @property
+    def variant_type(self):
+        if (self.start == self.end and len(self.ref) == len(self.alt) == 1
+        and self.ref != self.alt):
+            return "missense"
+        elif self.ref != "" and self.alt == "":
+            return "deletion"
+        elif (self.ref != "" and self.alt != ""
```
elenikiachaki: This will also capture `delins` that on the protein level look like insertions, like: `p.Ser12_Val13delinsSerAlaVal`, but since we handle variants on protein level, I assume that this is not a probelm.

cancermuts/core.py
```diff
@@ -261,14 +235,14 @@ def __init__(self, start, end, ref, alt, start_res,
...
     def variant_type(self):
         if (self.start == self.end and len(self.ref) == len(self.alt) == 1
-            and self.ref != self.alt):
-            return "substitution"
+        and self.ref != self.alt):
+            return "missense"
         elif self.ref != "" and self.alt == "":
             return "deletion"
-        elif self.ref == "" and self.alt != "":
+        elif (self.ref != "" and self.alt != ""
```
elenikiachaki: I decided to change the design for insertions, so that we can avoid the usage of `start_res `and `end_res `. For insertions, `ref` now contains the two WT flanking residues, and `alt `contains the same two flanking residues with the inserted sequence in between.

Example:
    `p.Ser12_Val13insAla`

represented internally as:
    start = 12
    end = 13
    ref = "SV"
    alt = "SAV"

cancermuts/core.py
elenikiachaki: Couldn't find a way to make good use of it. Tried with `._hgvs_parser.parse_p_posedit(item[2:])` but it is accepting both 1 and 3 letter notations (H39P and His39Pro)

cancermuts/core.py
elenikiachaki: but in the edge case of duplication in the C-terminal (`Ser27dup` when 27 is the last residue), this will not work since there is no right-flanking residue to create the object. Not sure if I should raise an error somewhere for this.

mtiberti: for insertions, can't it just be

```
start = 12
end = 12
ref = "S"
alt = "SAV"
```
so we anchor the insertion to the leftmost residue?

Technically speaking, I think that extra residues at N or C term would not be an insertion, they would be an [extension](https://hgvs-nomenclature.org/stable/recommendations/protein/extension/) - not sure what the priority is (in your example, would it be a duplication or an extension)? Do we need to support extensions at all?

mtiberti (changes requested):

cancermuts/properties.py
```diff
@@ -268,37 +252,35 @@ def __str__(self):
...
                self.disorder_propensity == other.disorder_propensity
 
     def __hash__(self):
-        return hash((self.source, self.disorder_propensity))
+        return hash((self.sources, self.disorder_propensity))
 
 class Structured(DisorderPropensity):
-    def __init__(self, position, sources):
+    def __init__(self, positions, sources):
         super(Ordered, self).__init__(name="Disorder Propensity",
```
elenikiachaki, line 263: Unrelated to PR but is Ordered here correct?

cancermuts/properties.py
```diff
@@ -268,37 +252,35 @@ def __str__(self):
...
-                                                 position=position,
+                                                 positions=positions,
                                                  sources=sources,
                                                  disorder_state='S')
 
 class Disordered(DisorderPropensity):
-    def __init__(self, position, sources):
+    def __init__(self, positions, sources):
         super(Ordered, self).__init__(name="Disorder Propensity",
```
elenikiachaki, line 270: Is ordered here correct?

mtiberti (approved):

PR #270, opened 2026-05-06
Modified table class and adjusted core.py

elenikiachaki:
Addresses comment #268 
## Summary

This PR updates the `Table` class to work with the refactored package where sequence annotations are stored as `SequenceProperty` objects in `sequence.properties`, and variants are managed through `VariantRegister`.

## Changes

- Added `Sequence.properties_at_position()` in `core.py`
- Updated `Table.to_dataframe()` to build residue-level rows from `Sequence.__iter__()`.
- Retrieved sequence properties from `sequence.properties_at_position()`.
- Updated mutation table integration to use `sequence.variants.get_variant_table()`.
- Added variant-specific columns:
  - `variant_hgvs`
  - `variant_start`
  - `variant_end`
  - `variant_type`
- Kept variants anchored to `variant_start`.
- Added metadata/property formatting helpers

### DIFF
--- a/cancermuts/core.py
+++ b/cancermuts/core.py
@@ -187,22 +187,26 @@ class Sequence(object):
 
     def properties_at_position(self, position, property_name=None):
         """
-        If property_name is provided, return a list of properties from that
-        property category.
-        If property_name is None, return a dictionary mapping
-        property categories to matching properties.
+        If property_name is provided, return a dict of properties mapped
+        to the property category.
+        If property_name is None, return all property categories with their
+        properties.
         """
 
         if property_name is not None:
+            if property_name not in sequence_properties_classes:
+                raise ValueError(f"Unknown property_name: {property_name}")
             if property_name not in self.properties:
-                return []
-            return [prop for prop in self.properties[property_name]
-                    if position in prop.positions]
+                properties_to_check = {property_name: []}
+            else:
+                properties_to_check = {property_name: self.properties[property_name]}
+        else:
+            properties_to_check = self.properties
 
         properties = {}
-        for this_property_name, props in self.properties.items():
+        for this_property_name, props in properties_to_check.items():
             matching_props = [prop for prop in props if position in prop.positions]
-            if len(matching_props) > 0:
+            if matching_props:
                 properties[this_property_name] = matching_props
         return properties
 
@@ -401,7 +405,7 @@ class VariantRegister:
         except KeyError:
             return False
 
-    def get_variant_table(self, variant_types=None, metadata=None, metadata_formatter=None):
+    def get_variant_table(self, variant_types=None, metadata=[], metadata_formatter=None):
         """
         Return a pandas DataFrame containing the ordered variants stored in the register.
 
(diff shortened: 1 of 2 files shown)


cancermuts/core.py
```diff
@@ -396,29 +417,50 @@ def get_variant_table(self, variant_types=None):
...
         """
-
+        if metadata is None:
+            metadata = []
         variants = self._sorted_variants()
         if variant_types is not None:
             if isinstance(variant_types, str):
                 variant_types = {variant_types}
             else:
                 variant_types = set(variant_types)
```
mtiberti, line 434: worth adding a check that the variant types are the ones we support?

cancermuts/core.py
```diff
@@ -396,29 +417,50 @@ def get_variant_table(self, variant_types=None):
...
+            variant_dictionary["variant_end"].append(v.end)
+            variant_dictionary["variant_ref"].append(v.ref)
+            variant_dictionary["variant_alt"].append(v.alt)
             variant_dictionary["variant_type"].append(v.variant_type)
-            variant_dictionary["variant"].append(str(v))
+
+            source_names = []
+            for source in v.sources:
+                if source.name not in source_names:
+                    source_names.append(source.name)
```
mtiberti: you could avoid this making `source_names` a `set` and then turning it into a list

mtiberti: e.g.

```
source names = list(set([v.sources]))
```

also, is this really necessary? Do we get cases in which the same source is repeated more than once?

cancermuts/core.py
```diff
@@ -185,6 +185,27 @@ def add_variant(self, var):
...
+        If property_name is None, return a dictionary mapping
+        property categories to matching properties.
+        """
+
+        if property_name is not None:
+            if property_name not in self.properties:
+                return []
+            return [prop for prop in self.properties[property_name]
+                    if position in prop.positions]
+
```
mtiberti, line 205: is it ok that this function returns a list with an argument set and a dict when it's not set?

in the first case, wouldn't it be more general to return a dict with one element (the property)?

cancermuts/core.py
```diff
@@ -396,29 +417,50 @@ def get_variant_table(self, variant_types=None):
         :obj:`pandas.DataFrame`
             DataFrame with one row per variant
         """
-
+        if metadata is None:
+            metadata = []
```
mtiberti: why not defaulting metadata to an empty list?

mtiberti (changes requested):

cancermuts/core.py
elenikiachaki: Yes, there are such cases, when clinvar has multiple variant ids for the same variant. Example on server: `/data/user/shared_projects/mavisp/SMAD2/cancermuts_idps_2026/metatable_pancancer_SMAD2.csv`
(So far, we have been storing multiple times "ClinVar" in sources.)

mtiberti: ok - this is a bug I believe, there's no reason we should have repeated sources in our table

mtiberti (approved):

PR #271, opened 2026-05-06
Isoform flag is used before exiting if there is more than one canonical id

pablosanchezizquierdo:
fix #269 Changed the order `datasources.py` so if isoform is provided and canoincal ids are more than one the program will not exit


cancermuts/datasources.py
```diff
@@ -179,17 +179,9 @@ def get_sequence(self, gene_id, upid=None, upac=None, isoform=None):
                         isoform_map[iso_id] = is_displayed
         except KeyError as e:
             raise ValueError(f"Missing expected field '{e.args[0]}' in ALTERNATIVE PRODUCTS for UniProt entry {this_upac}")
-
+
```
elenikiachaki: Could we remove these spaces?

PR #273, opened 2026-05-08
Throttle ClinVar requests

elenikiachaki:
Addresses issue #272 by adding a throttle for ClinVar/NCBI requests across all users in the servers. The timestamp of the last request is stored in a shared lock file in `/tmp`, limiting requests to 2 per second to reduce API overcrowding and repeated retry messages.


cancermuts/datasources.py
```diff
@@ -629,17 +630,61 @@ def _get_available_mutations(self, gene_id, metadata=[]):
         return mutations, out_metadata
 
 class ClinVar(DynamicSource, object):
+    _ncbi_rate_lock = threading.Lock()
```
mtiberti: in this PR, we are using two locking mechanisms at the same time, one at thread level (this one) and one at inter-process level

since we need to keep the second, I don't see why this one is actually necessary

cancermuts/datasources.py
```diff
@@ -629,17 +630,61 @@ def _get_available_mutations(self, gene_id, metadata=[]):
...
 
     # ClinVar Metadata
     _clinvar_supported_metadata = [
         'clinvar_germline_classification', 'clinvar_germline_condition', 'clinvar_germline_review_status', 'genomic_mutations',
         'clinvar_variant_id', 'genomic_coordinates', 'clinvar_oncogenicity_condition', 'clinvar_oncogenicity_classification',
         'clinvar_oncogenicity_review_status', 'clinvar_clinical_impact_condition', 'clinvar_clinical_impact_review_status', 'clinvar_clinical_impact_classification']
     @logger_init
-    def __init__(self):
+    def __init__(self, ncbi_requests_per_second=2.0, ncbi_rate_limit_lock_file="/tmp/cancermuts_ncbi_rate_limit.lock"):
```
mtiberti: I think the `requests_per_second` argument should default to an `int` (imho it's a bit weird that it's a float from an intuition prespective)

mtiberti: also can we check that this argument is not something silly, like <=0?

also, there should be a way to 1) disable the limit 2) disable the lock behavior (maybe setting arguments to None)?

cancermuts/datasources.py
```diff
@@ -629,17 +630,61 @@ def _get_available_mutations(self, gene_id, metadata=[]):
...
+            return open(path, "r+")
+        except FileNotFoundError:
+            try:
+                fd = os.open(path, os.O_RDWR | os.O_CREAT | os.O_EXCL, 0o666)
+                os.fchmod(fd, 0o666)
+                return os.fdopen(fd, "r+")
+            except FileExistsError:
+                return open(path, "r+")
+
+    def _ncbi_throttle(self):
```
mtiberti, line 665: tbh I'm not super into this approach of writing timestamp in a file, locking it, rereading it etc for every call to NCBI, since we are doing many calls frequently, and I feel like this adds an overhead as two concurrent processes will almost continuously try to read or write to this file.

Since the processing time between different calls that use the NCBI API is small, I'd rather have an approach by which a process generates a lock, than keeps it active until it's done all its requests, and then removes the lock so that another process can take over

of course this would require adding more safeguards internally to be sure that within a certain function/process we always wait before running consecutive calls

mtiberti (changes requested):

cancermuts/datasources.py
```diff
@@ -630,21 +629,26 @@ def _get_available_mutations(self, gene_id, metadata=[]):
...
-    _ncbi_rate_lock = threading.Lock()
 
     # ClinVar Metadata
     _clinvar_supported_metadata = [
         'clinvar_germline_classification', 'clinvar_germline_condition', 'clinvar_germline_review_status', 'genomic_mutations',
         'clinvar_variant_id', 'genomic_coordinates', 'clinvar_oncogenicity_condition', 'clinvar_oncogenicity_classification',
         'clinvar_oncogenicity_review_status', 'clinvar_clinical_impact_condition', 'clinvar_clinical_impact_review_status', 'clinvar_clinical_impact_classification']
     @logger_init
-    def __init__(self, ncbi_requests_per_second=2.0, ncbi_rate_limit_lock_file="/tmp/cancermuts_ncbi_rate_limit.lock"):
+    def __init__(self, ncbi_requests_per_second=8, ncbi_rate_limit_lock_file="/tmp/cancermuts_ncbi_rate_limit.lock", ncbi_api_key="c1e940e61e13bf4652922ded36f5fb2f2108"):
```
mtiberti: there should not be a default key here - definitely there shouldn't be your key on github, can you remove it? It could default to `None`

cancermuts/datasources.py
```diff
@@ -1509,48 +1549,57 @@ def _get_available_muts_and_md(self, sequence, clinvar_ids, metadata=[]):
...
-            raise TypeError(f"Missing 'refseq' alias for gene {sequence.gene_id}; cannot parse ClinVar variants.")
-        refseq = sequence.aliases["refseq"]
-        filter_missense_variants = f'({gene}[gene] AND ("missense variant"[molecular consequence] OR "SO:0001583"[molecular consequence]))'
-        mutation_type = "missense"
-        not_found_gene = []
-
+        lock_fh = None
+        if self._ncbi_rate_limit_lock_file is not None:
+            lock_fh = self._open_ncbi_lock_file()
+            fcntl.flock(lock_fh, fcntl.LOCK_EX)
```
mtiberti: can we add some logging about this event - so we know why the process is stalling? e.g. at warning level so it's usually printed out

mtiberti (changes requested):

cancermuts/datasources.py
mtiberti: also why did the default rate changed from 2 to 8?

elenikiachaki: That was my fault. In general:
- without the usage of a key, max reqests per second should be up to 10 per second
- with a key they can be up to 3 per second
but here I will set it back to 2.

mtiberti (approved):

PR #276, opened 2026-05-18
Updated docs and added method in core.py

elenikiachaki:
This PR addresses #274.

The printed outputs shown in` tutorial.md` were not consistent with the actual outputs I got when running `tutorial.py`, so I updated them. I’m not sure whether the documented outputs were outdated, came from a different protein/example, or if there is an issue in the current code and I am not getting the same results..

I also noticed that some print examples shown in `tutorial.md` were not actually printed by `tutorial.py`, so I added the corresponding print commands to the script.

Examples of outputs I could not reproduce and I am not sure where they are coming from:
```
>>> seq.positions[45].properties
{'ptm_glycosylation': <PositionProperty Glycosylation Site from GlyGen>}

>>> seq.positions[45].properties['ptm_glycosylation'].metadata
{'subtypes': ['N-GlcNAc', 'O-GalNAc']}

>>> seq.positions[10].properties['mobidb_disorder_propensity']
<StructuralDisorder, Structured, observed>
```


cancermuts/core.py
```diff
@@ -228,7 +254,7 @@ class ProteinVariant(object):
         metadata : :obj:`dict`
             Dictionary encoding variant-associated metadata.
     """
-
+    VARIANT_TYPES = {"missense", "deletion", "insertion", "delins"}
```
mtiberti: any reason this is uppercase? Also could it be something more explicit like `supported_variant_types`?

cancermuts/core.py
```diff
@@ -121,26 +121,30 @@ def __iter__(self):
...
-            self.log.debug("adding property %s to sequence of %s (%s)" % (str(prop), self.gene_id, add_type))
+        if positions is None:
+            positions = []
+        if not isinstance(positions, list):
+            raise TypeError("positions must be a list of 1-based residue numbers")
+
+        for position in positions:
+            if position not in self.sequence_numbering:
+                    raise ValueError(f"Property {property_name} is outside sequence bounds: "
+                                     f"position {position} for sequence of length {len(self.sequence)}")
```
mtiberti: could this be cleaner as a set operation?

```
if not set(positions).issubset(set(self.sequence_numbering))
```

Do we like that `positions` may contain repeated numbers? We could add a check

cancermuts/core.py
```diff
@@ -185,31 +189,58 @@ def add_variant(self, var):
...
+        else:
             if property_name not in sequence_properties_classes:
                 raise ValueError(f"Unknown property_name: {property_name}")
-            if property_name not in self.properties:
-                properties_to_check = {property_name: []}
-            else:
-                properties_to_check = {property_name: self.properties[property_name]}
-        else:
-            properties_to_check = self.properties
+            property_names = [property_name]
```
mtiberti: since we already have a setup for this, we could add support for a custom list of properties other than a single one?

cancermuts/core.py
```diff
@@ -121,26 +121,30 @@ def __iter__(self):
...
+            raise TypeError("positions must be a list of 1-based residue numbers")
+
+        for position in positions:
+            if position not in self.sequence_numbering:
+                    raise ValueError(f"Property {property_name} is outside sequence bounds: "
+                                     f"position {position} for sequence of length {len(self.sequence)}")
+        if property_name not in self.properties:
+            self.properties[property_name] = sequence_properties_classes[property_name]()
+
+        return self.properties[property_name].add(positions=positions, sources=sources, **metadata)
```
mtiberti: I don't understand what we are `return`ing here?

if `return` is not necessary, we can just remove it - so that it doesn't look like we mean to return something useful

cancermuts/properties.py
```diff
@@ -24,258 +24,185 @@
...
-        return self.__repr__()
-
-    def __eq__(self, other):
-        return self.sources == other.sources and \
-               self.disorder_propensity == other.disorder_propensity
+    category = "mobidb_disorder_propensity"
 
-    def __hash__(self):
-        return hash((self.sources, self.disorder_propensity))
+    def add(self, positions, sources=None, disorder_state=None, **metadata):
```
mtiberti: can we change the name of these `add` methods with something a bit more descriptive?

e.g. `add_entries` if we are adding entries

mtiberti (changes requested):

PR #282, opened 2026-06-09
Make ClinVar consistency cross-check optional

elenikiachaki:
PR that addresses #281 

It adds a `cross_check` flag to `ClinVar.add_mutations()`, defaulting to False.

The flag controls whether the additional ClinVar consistency check is performed.


pablosanchezizquierdo (approved):
Tested and it's fixing the issue by leaving cross-check optional and false as default

PR #283, opened 2026-06-11
Preserve glycosylation sources whith GlyGen and dbPTM annotations(#279)

drgbea:
fixes #279 

- datasources.py: fixed glycosylation annotation merging so that existing data sources are preserved when GlyGen and dbPTM annotate the same glycosylation site, independently of the order in which annotations are added
- properties.py: updated glycosylation subtype handling to retain all source information while preserving the most specific glycosylation subtype available for a given site


cancermuts/properties.py
```diff
@@ -188,7 +188,14 @@ def __init__(self, position, sources):
...
                                                     metadata={"subtypes": []}  )
 
-    def add_subtype(self,subtype):
+    def add_subtype(self, subtype):
+        base = subtype.split("-")[0]
+
+        self.metadata["subtypes"] = [
+            st for st in self.metadata["subtypes"]
+            if not (st.startswith(base + "-") and st != subtype)
+        ]
```
mtiberti: I'm not sure what this accomplishes

is it to avoid having repeated subtypes?

drgbea: The intention here was not to prevent duplicate subtypes (that is already handled by the check below), but to keep only the most specific glycosylation subtype available for a given site.
For example, if dbPTM annotates a site as `N-Gly` and GlyGen later annotates the same site as `N-GlcNAc`, the generic subtype is removed and only the more informative one is retained

mtiberti (dismissed):

mtiberti (approved):

PR #289, opened 2026-06-22
Addition of indel retrieval

elenikiachaki:
This PR addresses #280, adding support for retrieving and annotating **in-frame indels** in cancermuts.

**Main changes**
* Created the protein variant parser function to parse all supported variants types.
* Added a `variant_types` argument so users can request `"missense"`, `"deletion"`, `"insertion"` and/or `"delins"` variants.
* Added indel retrieval support for cBioPortal, COSMIC and ClinVar.
* Kept missense variants as the default, so existing usage remains unchanged.
* Updated genomic metadata/liftover handling so indel genomic annotations can be converted between assemblies.
* Extended gnomAD annotation to support indels when the genomic representation can be safely matched. Couldn't find a good solution for repeats and inversions and they are skipped.
* Added `tutorial_indels.py` using RAD51C as an example.
* Updated tutorial.md
* Skipped REVEL retrieval for variants types other than missense

**Notes**
- I followed the Ensembl VEP normalisation documentation  [VEP normalisation documentation ](https://www.ensembl.info/2018/06/22/cool-stuff-the-vep-can-do-normalisation/) for parsing the genomic mutations for gnomAD
- For indels, it may be useful to optionally show the protein HGVS string in the metatable, since `aa_position`, `ref_aa` and `alt_aa` can be harder to interpret


cancermuts/datasources.py
```diff
@@ -115,6 +117,93 @@ class StaticSource(Source, object):
     def __init__(self, *args, **kwargs):
         super(StaticSource, self).__init__(*args, **kwargs)
 
+def _protein_variant_parser(mutation, sequence, source):
```
mtiberti: I'm not sure I love having this function floating around - it breaks the consistency of the API for how we've developed it so far

so I propose this: we can have a `DynamicMutationSource` class that inherits from `DynamicSource` and defines this function as a class member, and then classes that need to use it have `DynamicMutationSource` as a base class

cancermuts/datasources.py
```diff
@@ -115,6 +117,93 @@ class StaticSource(Source, object):
...
+
```
mtiberti: given these definitions, would it possible for a variant to match more than one case for the same variant?

I was wondering if the order in which the variants are checked have any influence on their final classification

one idea could be to match all of them first and check if 2+ of them match. But then this is only necessary if we think it is possible that 2 or more can match

cancermuts/datasources.py
```diff
@@ -115,6 +117,93 @@ class StaticSource(Source, object):
...
+        start_ref, start, end_ref, end, alt = match.groups()
+        start = int(start)
+        end = int(end)
+        middle_ref = sequence.sequence[start:end - 1]
+        ref = start_ref + middle_ref + end_ref
+        return ProteinVariant(start=start, end=end, ref=ref, alt=alt, sources=[source])
+
+    raise ValueError(f"Unsupported protein mutation format: {mutation}")
+
+def _normalise_variant_types(variant_types):
```
mtiberti: same as previous re the location of this function

mtiberti: can we change the name?

I don't think this function normalizes things - it just removes duplicates

cancermuts/datasources.py
```diff
@@ -115,6 +117,93 @@ class StaticSource(Source, object):
...
+    elif isinstance(variant_types, str):
+        variant_types = (variant_types,)
+
+    variant_types = set(variant_types)
+    invalid_variant_types = variant_types - ProteinVariant.supported_variant_types
+
+    if invalid_variant_types:
+        raise ValueError(f"Invalid variant type(s): {invalid_variant_types}")
+
+    return variant_types
```
mtiberti: is returning a `set` here intentional?

cancermuts/datasources.py
```diff
@@ -367,9 +456,12 @@ def __init__(self, cancer_studies=None):
         self._get_cancer_types()
         self._get_cancer_studies(cancer_studies)
         self._get_molecular_profiles()
-
-        self._mut_regexp = '[A-Z][0-9]+[A-Z]$'
-        self._mut_prog = re.compile(self._mut_regexp)
+        self._mut_regexps = {"missense": re.compile(rf"^{aa}\d+{aa}$"),
+                             "deletion": re.compile(rf"^({aa}\d+del|{aa}\d+_{aa}\d+del)$"),
+                             "insertion": re.compile(rf"^({aa}\d+_{aa}\d+ins{aa}+|{aa}\d+(?:_{aa}\d+)?dup)$"),
+                             "delins": re.compile(rf"^({aa}\d+delins{aa}+|{aa}\d+_{aa}\d+delins{aa}+)$")}
```
mtiberti: since we use these in multiple places - probably worth having it in (and another reason for having) the `DynamicMutationSource`

cancermuts/datasources.py
```diff
@@ -367,9 +456,12 @@ def __init__(self, cancer_studies=None):
...
+        self._mut_prog = self._mut_regexps["missense"]
+        self._mut_regexp = self._mut_regexps["missense"].pattern
```
mtiberti: I'm not sure why this happens, if it's necessary can't we do so in the dictionary definition?

cancermuts/datasources.py
```diff
@@ -115,6 +117,93 @@ class StaticSource(Source, object):
...
+        end = int(end)
+        middle_ref = sequence.sequence[start:end - 1]
+        ref = start_ref + middle_ref + end_ref
+        return ProteinVariant(start=start, end=end, ref=ref, alt=alt, sources=[source])
+
+    raise ValueError(f"Unsupported protein mutation format: {mutation}")
+
+def _normalise_variant_types(variant_types):
+    if variant_types is None:
+        variant_types = ("missense",)
```
mtiberti: I think we should remove this case - see next comment when this function is used

cancermuts/datasources.py
```diff
@@ -379,9 +471,15 @@ def cancer_types(self):
...
-        _cBioPortal_supported_metadata = ['cancer_type', 'cancer_study', 'genomic_coordinates', 'genomic_mutations']
+    def _cbioportal_mutation_type(self, mutation):
+        for variant_type, pattern in self._mut_regexps.items():
+            if pattern.fullmatch(mutation):
+                return variant_type
+        return None
 
+    def add_mutations(self, sequence, metadata=[], variant_types=("missense",)):
+        _cBioPortal_supported_metadata = ['cancer_type', 'cancer_study', 'genomic_coordinates', 'genomic_mutations']
+        variant_types = _normalise_variant_types(variant_types)
```
mtiberti: I think that if `variant_types` is `None` or empty or any non-supported type we should raise an exception - I don't think `None` should equal missense mutations. However, I like that missense is the default for this function

cancermuts/datasources.py
```diff
@@ -394,42 +492,36 @@ def add_mutations(self, sequence, metadata=[]):
...
                         mutation_obj.metadata[md].append(this_md)
 
-            self.log.debug("adding mutation %s" % str(mutation_obj))
-            sequence.add_variant(mutation_obj)
+            try:
+                self.log.debug("adding mutation %s" % str(mutation_obj))
+                sequence.add_variant(mutation_obj)
+            except ValueError as e:
+                self.log.warning(f"mutation {mutation} failed wild-type validation "
+                                 f"and will be skipped: {e}")
```
mtiberti: I think we could be a bit more explicit with this error message - what is "wild-type validation" for a non-developer?

cancermuts/datasources.py
```diff
@@ -499,14 +591,23 @@ def _get_mutation_data(self, gene_id):
         df = pd.DataFrame(dict(
             [ (attr, [ getattr(entry, attr) for entry in result ]) for attr in dir(result[0]) ]))
 
-        df = df[df['mutationType'] == 'Missense_Mutation']
-        df = df[df['proteinChange'].str.match(self._mut_regexp, case=True)]
+        variant_types = _normalise_variant_types(variant_types)
```
mtiberti: this is already done in `add_mutations` and shouldn't be necessary here

I don't think it's the right place to sanitize arguments, closer to where they are given first (as in `add_mutations`) is a much better place

cancermuts/datasources.py
```diff
@@ -499,14 +591,23 @@ def _get_mutation_data(self, gene_id):
...
 
-        return df
+        allowed_cbioportal_types = {"missense": "Missense_Mutation",
+                                    "deletion": "In_Frame_Del",
+                                    "delins": "In_Frame_Del",
+                                    "insertion": "In_Frame_Ins"}
+
+        keep = pd.Series(False, index=df.index)
+        for variant_type in variant_types:
+            keep |= ((df["mutationType"] == allowed_cbioportal_types[variant_type]) & df["proteinChange"].apply(lambda mutation: self._cbioportal_mutation_type(mutation) == variant_type))
```
mtiberti: bit cryptic but also cool, I'd keep it, just add a comment please :)

cancermuts/datasources.py
```diff
@@ -499,14 +591,23 @@ def _get_mutation_data(self, gene_id):
...
 
-    def _get_available_mutations(self, gene_id, metadata=[]):
+        return df[keep]
 
-        mutation_data = self._get_mutation_data(gene_id)
+    def _get_available_mutations(self, gene_id, metadata=[], variant_types=("missense",)):
+
+        mutation_data = self._get_mutation_data(gene_id, variant_types=variant_types)
```
mtiberti: I'm not sure why these two functions need to be split like this - they aren't reused

cancermuts/datasources.py
```diff
@@ -609,20 +710,29 @@ def _get_available_mutations(self, gene_id, metadata=[]):
...
-                            gm_fmt = f"{row['chr']}:g.{row['startPosition']}_{row['endPosition']}delins{row['variantAllele']}"
+                        elif  (row['referenceAllele'] != row['variantAllele'] and len(row['referenceAllele']) == len(row['variantAllele']) == 1
+                        and row['startPosition'] == row['endPosition']):
+                                gm_fmt = f"{row['chr']}:g.{row['startPosition']}{row['referenceAllele']}>{row['variantAllele']}"
+                                gm = ['hg19', gm_fmt]
+                        elif row['referenceAllele'] != '-' and row['variantAllele'] != '-' and row['referenceAllele'] != row['variantAllele']:
+                            if row['startPosition'] == row['endPosition']:
+                                gm_fmt = f"{row['chr']}:g.{row['startPosition']}delins{row['variantAllele']}"
+                            else:
+                                gm_fmt = f"{row['chr']}:g.{row['startPosition']}_{row['endPosition']}delins{row['variantAllele']}"
```
mtiberti: let's give this a thought - is there any better way to handle it?

It feels a bit spaghetti-esque. I was wondering if we can do this at `metadata.GenomicMutation` level since there's some code there that does a similar job already

cancermuts/datasources.py
```diff
@@ -1493,21 +1614,29 @@ def _get_available_muts_and_md(self, sequence, clinvar_ids, metadata=[], cross_c
...
-                    self.log.warning(
-                        f"ClinVar ID {clinvar_id}: genomic annotation '{gm.definition}' for {gm.genome_build} "
-                        f"is not in the expected for missense variant format and will not be added to genomic metadata")
+                supported_genomic_annotation = (gm.is_snv
+                                               or gm.is_insdel
+                                               or gm.is_inversion
+                                               or gm.is_deletion
+                                               or gm.is_insertion
+                                               or gm.is_duplication
+                                               or gm.is_repeat)
```
mtiberti: a reason more we should think about a refactoring of `metadata.GenomicMutation` - it was ok when we were supporting 1/2 types but it's not very robust as it is

cancermuts/datasources.py
```diff
@@ -1493,21 +1614,29 @@ def _get_available_muts_and_md(self, sequence, clinvar_ids, metadata=[], cross_c
...
+
+                if gm.chr is None or not supported_genomic_annotation:
```
mtiberti: do we really get a `None` chromosome?

cancermuts/datasources.py
```diff
@@ -1493,21 +1614,29 @@ def _get_available_muts_and_md(self, sequence, clinvar_ids, metadata=[], cross_c
...
-                     gm.coord if gm.is_snv else gm.coord_start,
-                     gm.coord if gm.is_snv else gm.coord_end,
-                     gm.ref if gm.is_snv else gm.substitution]
-                    for gm in valid_gms])
+                    out_metadata["genomic_coordinates"].append([[gm.genome_build,
+                                                                 gm.chr,
+                                                                 gm.coord if gm.is_snv else gm.coord_start,
+                                                                 gm.coord if gm.is_snv else gm.coord_end,
+                                                                 gm.ref if gm.is_snv else None]
+                                                                 for gm in valid_gms])
```
mtiberti, line 1701: see above re refactoring

cancermuts/datasources.py
```diff
@@ -3043,24 +3194,47 @@ def _get_metadata(self, mutation, gene_id, md_type):
...
+            if variant.is_snv or variant.is_insdel:
+                v_str = variant.get_value_str(fmt='gnomad')
+
+                if variant.is_snv:
+                    this_df = data[data['variant_id'] == v_str]
+                elif variant.is_insdel:
+                    this_df = data[data['edited_variant_id'] == v_str]
+
+            elif variant.is_insertion:
+                v_str = variant.definition
```
mtiberti: should we not specialize `GenomicVariant.get_value_str()` to return a format that makes sense to use for gnomad instead of doing all of the following?

cancermuts/metadata.py
mtiberti, line 139: again - this is a bit convoluted as it is; let's give a thought if a refactoring would be better

mtiberti (changes requested):

cancermuts/datasources.py
elenikiachaki: In theory, with the current fullmatch-based patterns, the variant classes should be mutually exclusive.

cancermuts/datasources.py
```diff
@@ -62,6 +62,8 @@
 if sys.version_info[0] >= 3:
     unicode = str
 
+aa = r"[ACDEFGHIKLMNPQRSTVWY]"
```
mtiberti: since this is only used in `DynamicMutationSource, can we define it there?

cancermuts/datasources.py
```diff
@@ -107,6 +109,110 @@ class DynamicSource(Source, object):
...
+            end = int(end)
+            middle_ref = sequence.sequence[start:end - 1]
+            ref = start_ref + middle_ref + end_ref
+            return ProteinVariant(start=start, end=end, ref=ref, alt=alt, sources=[self])
+
+        raise ValueError(f"Unsupported protein mutation format: {mutation}")
+
+    def _validate_variant_types(self, variant_types):
+        if variant_types is None:
+            raise ValueError("variant_types cannot be None")
```
mtiberti: is this really necessary?

I'm just thinking that in this case the function could just return a "non valid" answer

cancermuts/datasources.py
```diff
@@ -107,6 +109,110 @@ class DynamicSource(Source, object):
...
+        elif isinstance(variant_types, str):
+            variant_types = (variant_types,)
+
+        variant_types = tuple(dict.fromkeys(variant_types))
+        if len(variant_types) == 0:
+            raise ValueError("variant_types cannot be empty")
```
mtiberti: as per previous comment - I think this is necessary if we need to distinguish the two cases, otherwise it could just be negative evaluation

cancermuts/datasources.py
```diff
@@ -107,6 +109,110 @@ class DynamicSource(Source, object):
...
+            start_ref, start, end_ref, end, alt = match.groups()
+            start = int(start)
+            end = int(end)
+            middle_ref = sequence.sequence[start:end - 1]
+            ref = start_ref + middle_ref + end_ref
+            return ProteinVariant(start=start, end=end, ref=ref, alt=alt, sources=[self])
+
+        raise ValueError(f"Unsupported protein mutation format: {mutation}")
+
+    def _validate_variant_types(self, variant_types):
```
mtiberti: this function is a bit unconventional to me

I would expect a validation function to return `True`/`False` rather than its arguments (e.g. to allow patterns like `if self._validate_variant_types(...)` (and then maybe it should be renamed to something like `self._valid_variant_types`, so we could do `if self._valid_variant_types(...)`. I think this would be a more conventional pattern, since the function has its purpose to validate and so I would expect its return values to be the default, rather than raise an exception in the negative case

cancermuts/datasources.py
```diff
@@ -379,9 +483,30 @@ def cancer_types(self):
...
+
+        variant = f"{chrom},{start},{end},{ref},{alt}"
+        try:
+            hgvsg = rq.get(f"{base_url}/annotation/genomic/{variant}").json().get("hgvsg")
+        except Exception as e:
+            self.log.error(f"failed retrieving HGVSg from Genome Nexus for {variant}")
+            hgvsg = None
 
+        self._hgvsg_cache[key] = hgvsg
+        return hgvsg
```
mtiberti: I don't love depending on an yet another external web service, but then again I couldn't find a "local" solution that would really be satisfactory. We could go with a combination of manual handling and hgvs - I played with ChatGPT a bit and this is what I came up with:

```
def vcf_like_to_hgvsg(chrom, pos, ref, alt, chrom_map=GRCH38_CHROM_TO_REFSEQ):
    chrom = chrom.removeprefix("chr")
    ac = chrom_map[chrom]
    pos = int(pos)
    ref = ref.upper()
    alt = alt.upper()

    # SNV
    if len(ref) == 1 and len(alt) == 1:
        raw = f"{ac}:g.{pos}{ref}>{alt}"

    # VCF-style deletion, e.g. pos=100, ref=ATC, alt=A
    elif len(ref) > len(alt) and ref.startswith(alt):
        deleted = ref[len(alt):]
        start = pos + len(alt)
        end = start + len(deleted) - 1
        raw = (
            f"{ac}:g.{start}del{deleted}"
            if start == end
            else f"{ac}:g.{start}_{end}del{deleted}"
        )

    # VCF-style insertion, e.g. pos=100, ref=A, alt=ATC
    elif len(alt) > len(ref) and alt.startswith(ref):
        inserted = alt[len(ref):]
        raw = f"{ac}:g.{pos}_{pos + 1}ins{inserted}"

    # MNV / complex replacement
    else:
        end = pos + len(ref) - 1
        raw = f"{ac}:g.{pos}_{end}delins{alt}"

    var = hp.parse_hgvs_variant(raw)
    var = hn.normalize(var)
    hv.validate(var)

    return str(var)
```

I can't tell how robust this would be but maybe worth a try?

Otherwise I think we need to change this into its own data source, so that it's explicit and we can properly credit it in the docs and such

cancermuts/datasources.py
mtiberti: feels like this should still be the case

cancermuts/metadata.py
```diff
@@ -159,132 +163,146 @@ def __init__(self, source, genome_build, definition):
...
-            self.coord_end = None
-            self.ref = None
-            self.alt = None
-            self.substitution = None
-            self.is_snv = False
-            self.is_insdel = False
-            self.is_inversion = False
+            return f"{self.start}_{self.end}"
+
+    def _parse_definition(self, definition):
```
mtiberti: this makes me believe we would be better off with the specialized per-mutation-type derived class we talked about

cancermuts/metadata.py
```diff
@@ -159,132 +163,146 @@ def __init__(self, source, genome_build, definition):
...
+        converted_coords_end = lo.convert_coordinate('chr%s' % self.chr, int(self.end))
+
+        if len(converted_coords_start) != 1:
+            raise TypeError("Could not convert genomic coordinates (liftOver returned %d coords)" % len(converted_coords_start))
+        if len(converted_coords_end) != 1:
+            raise TypeError("Could not convert genomic coordinates (liftOver returned %d coords)" % len(converted_coords_end))
+
+        converted_start = converted_coords_start[0][1]
+        converted_end = converted_coords_end[0][1]
+
```
mtiberti, line 239: same as previous comments re all the if/else - sign we need to specialize this into multiple classes

mtiberti (changes requested):
the patterns we have here re all the different mutation types in `metadata.GenomicMutation` make me think that we would be better off with one derived class per mutation type so that each would be defined more naturally. I think that this would be a better pattern than all the `if ... else` we have now, those are usually a sign that different types would be better off as being separate classes

cancermuts/datasources.py
elenikiachaki: This is how mutations are represented in cBioportal:
Category | Protein change | Chr | Start | End | Ref | Alt | Build
-- | -- | -- | -- | -- | -- | -- | --
Deletion | E746_A750del | 7 | 55242466 | 55242480 | GAATTAAGAGAAGCA | - | GRCh37
Insertion | N771_H773dup | 7 | 55249010 | 55249011 | - | ACAACCCCC | GRCh37
Complex replacement | T302H | 7 | 55223536 | 55223539 | GACA | ACAC | GRCh37

elenikiachaki: It is MAF format according to cbioportal documentation (https://docs.cbioportal.org/file-formats/#mutation-data):

cBioPortal mutation data file format

The cBioPortal mutation data file format recognized by the portal has:

- 32 columns from the [TCGA MAF format](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/).
- 1 column with the amino acid change.
- 4 columns with information on reference and variant allele counts in tumor and normal samples

Therefore, I adjusted the parsing you suggested.

elenikiachaki: On the other hand, gnomAD uses VCF representation, where REF and ALT cannot be empty. Insertions and deletions are therefore anchored to the left-flanking nucleotide. I am pointing this out because cBioPortal and gnomAD representations may look equivalent, but they are not.

cancermuts/datasources.py
```diff
@@ -197,21 +199,21 @@ def _protein_variant_parser(self, mutation, sequence):
...
-            raise ValueError("variant_types cannot be None")
-        elif isinstance(variant_types, str):
+            return False
+        if isinstance(variant_types, str):
             variant_types = (variant_types,)
+        try:
+            variant_types = tuple(variant_types)
+        except TypeError:
+            return False
+        return (bool(variant_types) and set(variant_types).issubset(ProteinVariant.supported_variant_types))
```
mtiberti: what is the `bool()` here for, is it to check if the list is empty? If so it would be more Pythonic to use `len(variant_types) != 0`

cancermuts/datasources.py
```diff
@@ -473,7 +475,10 @@ def __init__(self, cancer_studies=None):
         self._get_cancer_types()
         self._get_cancer_studies(cancer_studies)
         self._get_molecular_profiles()
-        self._hgvsg_cache = {}
+        self.hdp = uta.connect()
+        self.hp = parser.Parser()
+        self.hn = normalizer.Normalizer(self.hdp)
+        self.hv = validator.Validator(self.hdp)
```
mtiberti: I don't think these should be part of the API - if so their name should conventionally start with `_`. Also, can you please make these variable names more explicit? For a third party it's really unclear with "hdp" "hp" etc mean

cancermuts/datasources.py
```diff
@@ -483,30 +488,44 @@ def cancer_types(self):
...
+        elif ref == "-" and alt != "-":
+            raw = f"{ac}:g.{start}_{end}ins{alt}"
+
+        # MNV / complex replacement
+        else:
+            raw = (f"{ac}:g.{start}delins{alt}" if start == end else f"{ac}:g.{start}_{end}delins{alt}")
+
+        var = self.hp.parse_hgvs_variant(raw)
+        var = self.hn.normalize(var)
+        self.hv.validate(var)
```
mtiberti: how do we handle the case in which the variant is not validated?

cancermuts/datasources.py
```diff
@@ -58,12 +58,14 @@
 from io import StringIO
 import xmltodict
 import fcntl
+from bioutils.assemblies import make_name_ac_map
+from hgvs import parser, normalizer, validator
+from hgvs.dataproviders import uta
```
mtiberti, line 63: we also need to add this packages to the setup system as requirements and check if there are parts of the documentation where they need to be cited

cancermuts/datasources.py
```diff
@@ -659,6 +678,9 @@ def _get_available_mutations(self, gene_id, metadata=[], variant_types=("missens
             out_metadata['genomic_mutations'] = []
             do_genomic_mutations = True
 
+            chrom_maps = {"GRCh37": make_name_ac_map("GRCh37.p13", primary_only=True),
+                          "GRCh38": make_name_ac_map("GRCh38.p14", primary_only=True),}
```
mtiberti: nice, I didn't know this was a thing!

cancermuts/datasources.py
```diff
@@ -3137,10 +3176,29 @@ def _get_popmax_genome_allele_freq(self, mutation, gene_id):
...
+        pos = int(row["pos"])
+        ref = str(row["ref"]).upper()
+        alt = str(row["alt"]).upper()
+
+        # SNV
+        if len(ref) == 1 and len(alt) == 1:
+            return f"{chrom}-{pos}-{ref}-{alt}"
+
+        # Insertion: REF is the anchor and ALT is anchor + inserted sequence
+        if len(ref) == 1 and alt.startswith(ref):
```
mtiberti: maybe we could add `and len(alt) > 1`?

I understand it's kinda implied, since we will most likely not have cases of synonymous mutations, but you never know

also because these `if`s are not `elif`s

cancermuts/metadata.py
```diff
@@ -140,21 +140,16 @@ class GenomicMutation(Metadata):
...
-            return 'X'
-        elif chromosome == '24':
-            return 'Y'
-        return chromosome
+    def __new__(cls, source, genome_build, definition):
+        if cls is GenomicMutation and isinstance(definition, str):
+            for mutation_class in cls._mutation_classes:
+                if mutation_class._pattern.fullmatch(definition):
+                    return object.__new__(mutation_class)
+        return object.__new__(cls)
```
mtiberti, line 158: nice class factory approach!

maybe some comments would be helpful since it's not a very commonly used pattern

cancermuts/metadata.py
```diff
@@ -168,77 +163,41 @@ def __init__(self, source, genome_build, definition):
...
-        else:
-            return f"{self.start}_{self.end}"
-
-    def _parse_definition(self, definition):
-        match = self._mut_snv_prog.match(definition)
-        if match:
-            self._set_common_fields(match.groupdict(), "snv")
-            self.definition = f"{self.chr}:g.{self.start}{self.ref}>{self.alt}"
+        if type(self) is GenomicMutation:
+            self.log.info(f"unsupported genomic mutation format: {definition}")
```
mtiberti: I think this could be a warning

cancermuts/metadata.py
```diff
@@ -326,6 +268,113 @@ def __eq__(self, other):
...
+        return f"{chrom}:g.{self._position_string(start, end)}{self.alt}"
+
+GenomicMutation._mutation_classes = (GenomicSNV,
+                                     GenomicDelins,
+                                     GenomicInversion,
+                                     GenomicInsertion,
+                                     GenomicDuplication,
+                                     GenomicDeletion,
+                                     GenomicRepeat)
+
```
mtiberti, line 499: I think this is overall much nicer :)

docs/metatable_indels.csv
```diff
@@ -182,7 +182,7 @@
 180,181,Q,QHLQL,ClinVar,,,,,,,,,,,,,,,,"hg19,17:56774185-56774193, hg38,17:58696824-58696832","hg19,17:g.56774185_56774193dup, hg38,17:g.58696824_58696832dup",,,,,,,,3312384,Uncertain significance,Hereditary cancer-predisposing syndrome,1,,,,,,
 181,182,L,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 182,183,I,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
-183,184,A,,cBioPortal,,,,,,,,,,,,,,es_dsrct_msk_2023,Ewing Sarcoma,"hg19,17:56774198-56774200","hg19,17:g.56774198_56774200del",,,,,,,,,,,,,,,,,
+183,184,A,,cBioPortal,,,,,,,,,,,,,,es_dsrct_msk_2023,Ewing Sarcoma,"hg19,17:56774198-56774200","hg19,17:g.56774200_56774202del",,,,,,,,,,,,,,,,,
```
mtiberti: any reason this has changed? Do we know if this corresponds to a change in cBioPortal?

mtiberti (changes requested):

cancermuts/datasources.py
elenikiachaki: If the variant is not validated, Validator.validate() raises an exception that is caught when the method is called here (https://hgvs.readthedocs.io/en/stable/examples/using-hgvs.html?):
 
```
if do_genomic_mutations:
    chrom_map = chrom_maps[row["ncbiBuild"]]
    try:
        hgvsg = self._cbioportal_to_hgvsg(chrom=chrom, start=start, end=end, ref=ref, alt=alt, chrom_map=chrom_map)
        gm = [genome_build, hgvsg]

    except Exception as e:
        self.log.warning(f"Could not convert cBioPortal variant "
                         f"{chrom}:{start}-{end} {ref}>{alt} to HGVSg: {e}")
        gm = None
```
Let me know if we should change this behavior

docs/metatable_indels.csv
elenikiachaki: This happens because of the hgvs.normaliser. I added the following print statements to test it:
```
  var = self._hgvs_parser.parse_hgvs_variant(raw)
  var = self._hgvs_normalizer.normalize(var)
  self._hgvs_validator.validate(var)

  print("Raw:", raw)
  print("Normalized:", var)
```
and I got:
Normalized: NC_000017.10:g.56774200_56774202del
Raw: NC_000017.10:g.56774175_56774190delinsA

I think it happens because there are nucleotide repeats in that locus and the normaliser choses the repeat closest to the 3' as the one to be deleted (https://hgvs.readthedocs.io/en/stable/examples/using-hgvs.html?).

docs/metatable_indels.csv
```diff
@@ -184,7 +184,7 @@
 182,183,I,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 183,184,A,,cBioPortal,p.Ala184del,,,,,,,,,,,,,,es_dsrct_msk_2023,Ewing Sarcoma,"hg19,17:56774198-56774200","hg19,17:g.56774200_56774202del",,,,,,,,,,,,,,,,,
 184,185,E,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
-185,186,K,KKH,ClinVar,p.Lys186_His187insLys,,,,,,,,,,,,,,,,"hg19,17:56774205-56774207, hg38,17:58696844-58696846","hg19,17:g.56774205_56774207dup, hg38,17:g.58696844_58696846dup",,,,,,,,482173,Uncertain significance,"Hereditary cancer-predisposing syndrome;not provided;Breast-ovarian cancer, familial, susceptibility to, 3;Fanconi anemia complementation group O",2,,,,,,
+185,186,K,KKH,ClinVar,p.Lys186_His187insLys,,,,,,,,,,,,,,,,"hg19,17:56774205-56774207, hg38,17:58696844-58696846","hg19,17:g.56774205_56774207dup, hg38,17:g.58696844_58696846dup",,,,0.0000159067229765,0.0000318511912346,0.0002174622159400,0.0006410256410256,482173,Uncertain significance,"Hereditary cancer-predisposing syndrome;not provided;Breast-ovarian cancer, familial, susceptibility to, 3;Fanconi anemia complementation group O",2,,,,,,
```
elenikiachaki, line 187: for these variants we are now able to retrieve the gnomad frequency because of the left normalisation of their vcf

docs/metatable_indels.csv
```diff
@@ -374,4 +374,4 @@
 372,373,E,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 373,374,E,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
 374,375,E,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
-375,376,L,,ClinVar,p.Leu376del,,,,,,,,,,,,,,,,"hg19,17:56811579-56811581, hg38,17:58734218-58734220","hg19,17:g.56811579_56811581del, hg38,17:g.58734218_58734220del",,,,,,,,490119,Uncertain significance,Hereditary cancer-predisposing syndrome;not provided;Fanconi anemia complementation group O,2,,,,,,
+375,376,L,,ClinVar,p.Leu376del,,,,,,,,,,,,,,,,"hg19,17:56811579-56811581, hg38,17:58734218-58734220","hg19,17:g.56811579_56811581del, hg38,17:g.58734218_58734220del",,,,0.0000040544595000,,0.0000000000000000,,490119,Uncertain significance,Hereditary cancer-predisposing syndrome;not provided;Fanconi anemia complementation group O,2,,,,,,
```
elenikiachaki, line 377: Same as previous

cancermuts/datasources.py
```diff
@@ -3073,6 +3093,9 @@ class gnomAD(DynamicSource, object):
     _exome_genome_support = ['2.1', '2.1_controls', '2.1_non-neuro', '2.1_non-cancer', '2.1_non-topmed']
     _genome_support = ['3']
 
+    _reference_fastas = {"GRCh37": "/data/databases/genome_annotation/hg19.fa",
+                         "GRCh38": "/data/databases/genome_annotation/hg38.fa"}
```
mtiberti: this cannot be hardcoded - it needs to be set at constructor level, with user-definable arguments. 

it also needs to be clear in the documentation that these are necessary, what they are and where they come from

cancermuts/datasources.py
```diff
@@ -3240,37 +3303,49 @@ def _get_metadata(self, mutation, gene_id, md_type):
...
+        for variant in mutation.metadata["genomic_mutations"]:
+            if (not isinstance(variant, GenomicMutation)
+                or variant.mutation_type is None):
                 continue
             try:
                 variant = variant.as_assembly(ref_assembly)
-            except TypeError as e:
-                self.log.error(str(e))
+            except Exception as e:  
+                self.log.warning(f"Could not convert genomic variant {variant.definition} to {ref_assembly}: {e}")
```
mtiberti, line 3326: here having just `{e}` as the text is a bit misleading, it only prints the argument of the exception, i.e. if it's a `KeyError` it would just be the key. Is it possible to e.g. distinguish the two errors, one in case liftover just couldn't convert assembly or other processing issues? It would be useful for troubleshooting

cancermuts/datasources.py
```diff
@@ -3240,37 +3303,49 @@ def _get_metadata(self, mutation, gene_id, md_type):
...
 
-            v_str = variant.get_value_str(fmt="gnomad")
-            if v_str is None:
-                self.log.info(f"variant type not supported by gnomAD:{variant.definition}")
+            reference_sequence = None
+            if variant.requires_reference_sequence:
+                try:
+                    reference_sequence = self._get_reference_sequence(variant, ref_assembly,)
+                except Exception as e:
+                    self.log.warning(f"Could not retrieve reference sequence for {variant.definition}: {e}")
```
mtiberti: in general terms, we should try and avoid catch-all `except Exception` because they are very good at hiding bugs

mtiberti: this comment is relevant for this whole function definition (or more)

cancermuts/metadata.py
```diff
@@ -280,13 +286,14 @@ def _set_specific_fields(self, tokens):
     def _build_definition(self, chrom, start, end):
         return f"{chrom}:g.{start}{self.ref}>{self.alt}"
 
-    def get_value_str(self, fmt="csv"):
+    def get_value_str(self, fmt="csv", reference_sequence=None, normalize_vcf=None):
```
mtiberti, line 289: this and all the following similarly-defined function have this `normalize_vcf` argument that is not used anywhere, for what I can tell, so I'm not sure what its point is

cancermuts/metadata.py
```diff
@@ -340,22 +394,53 @@ def _set_specific_fields(self, tokens):
 
     def _build_definition(self, chrom, start, end):
         return f"{chrom}:g.{start}_{end}ins{self.alt}"
-
-    def get_value_str(self, fmt="csv"):
+
```
mtiberti: whitespace to clean up

docs/install.md
```diff
@@ -240,4 +243,25 @@ The dataset integrates information from multiple resources curated in GlyGen. Wh
...
+local GRCh37 and GRCh38 reference FASTA files. Cancermuts uses these
+references to reconstruct and left-normalize VCF alleles through the
+bcftools functionality provided by `pysam`.
+
+The FASTA paths must be configured in the gnomAD datasource:
+
+```python
+_reference_fastas = {"GRCh37": "/path/to/hg19.fa",
+                     "GRCh38": "/path/to/hg38.fa"}
+  ```
```
mtiberti: see my comment on this in the code

mtiberti (changes requested):

cancermuts/metadata.py
elenikiachaki: `normalize_vcf `is used by the non-SNV classes to normalize the VCF allele. It is passed from the gnomAD in `datasources.py` here:

```
variant.get_value_str(fmt="gnomad", reference_sequence=reference_sequence, normalize_vcf=self._normalize_vcf_allele)
```

cancermuts/datasources.py
```diff
@@ -3094,17 +3094,15 @@ class gnomAD(DynamicSource, object):
     _exome_genome_support = ['2.1', '2.1_controls', '2.1_non-neuro', '2.1_non-cancer', '2.1_non-topmed']
     _genome_support = ['3']
 
-    _reference_fastas = {"GRCh37": "/data/databases/genome_annotation/hg19.fa",
-                         "GRCh38": "/data/databases/genome_annotation/hg38.fa"}
-
     @logger_init
-    def __init__(self, version='2.1'):
+    def __init__(self, version='2.1', reference_fasta=None):
```
mtiberti: can we rename `reference_fasta` to something a bit more explicit? e.g. `genome_fasta` or `refrence_genome_fasta` so it's more clear what it is about

mtiberti (changes requested):

mtiberti (approved):

PR #296, opened 2026-07-30
Fix UniProt alias error handling

### LINKED ISSUE #294: `cancermuts.datasources.UniProt._get_aliases` has poor error handling
in particular, the function just returns `None` if any of the requested aliases is not returned

it should behave more constructively; either return the full dictionary with a `None` entry or `raise` an exception depending on what would work best in-context

pablosanchezizquierdo:
fix #294 

This PR improves the error handling in `UniProt._get_aliases()`.

### Changes

- return a complete alias dictionary instead of returning `None` when an alias cannot be resolved
- return `None` for unresolved aliases while continuing to process the remaining requested aliases
- handle invalid responses from the UniProt mapping service more robustly

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -463,7 +463,7 @@ class UniProt(DynamicSource, object):
 
             if t_keyword is not None:
                 self.log.info('using extracted keyword %s to parse results' % t_keyword)
-                out[t] = results['to'].get(t_keyword)
+                out[t] = results['to'][t_keyword]
             else:
                 out[t] = results['to']
 
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -247,6 +247,8 @@ class UniProt(DynamicSource, object):
         elif upid is not None:
             self.log.info("UniProt AC will be mapped from UniProt ID")
             this_upac = self._get_aliases(upid, ['UniProtKB_primaryAccession'])['UniProtKB_primaryAccession']
+            if this_upac is None:
+                raise ValueError(f"Could not resolve UniProt primary accession for {upid}")
         else:
             self.log.info("retrieving UniProt ID for human gene %s" % gene_id)
             try:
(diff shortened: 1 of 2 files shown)


cancermuts/datasources.py
```diff
@@ -446,7 +458,7 @@ def _get_aliases(self, gene_id, to, fr='UniProtKB_AC-ID'):
 
             if t_keyword is not None:
                 self.log.info('using extracted keyword %s to parse results' % t_keyword)
-                out[t] = results['to'][t_keyword]
+                out[t] = results['to'].get(t_keyword)
```
mtiberti: why this change?

cancermuts/datasources.py
```diff
@@ -429,13 +429,25 @@ def _get_aliases(self, gene_id, to, fr='UniProtKB_AC-ID'):
...
+            responses = mapping_result.get('results', [])
 
             if len(responses) == 0:
                 self.log.warning(f"No {t} found for {gene_id}")
-                return None
+                out[t] = None
+                continue
 
             if len(responses) > 1:
                 self.log.warning(f"Multiple {t} found for {gene_id}. Found {t}: {', '.join(response['to'] for response in responses)}. No {t} will be assigned.")
```
mtiberti, line 458: in this case, should we also set `out[t]` as per the previous check?

mtiberti (changes requested):
this is fine, but I think we should also handle this case in the calling function.

What happens if the conversion fails and this is `None`?

I think we should check and raise an Exception in the calling function e.g. for `ValueError` unless we have a way around it

cancermuts/datasources.py
pablosanchezizquierdo: it was to avoid KeyError but if we handle the error in get_sequence I can revert it

cancermuts/datasources.py
pablosanchezizquierdo: you mean the out[t] = None?
 its already there

cancermuts/datasources.py
```diff
@@ -264,9 +266,12 @@ def get_sequence(self, gene_id, upid=None, upac=None, isoform=None):
                 self.log.info("will use Uniprot ID %s" % this_upid)
 
             this_upac = self._get_aliases(this_upid, ['UniProtKB_primaryAccession'])['UniProtKB_primaryAccession']
-
+            if this_upac is None:
+                raise ValueError(f"Could not resolve UniProt primary accession for {this_upid}")
         if upid is None:
             this_upid = self._get_aliases(this_upac, ['UniProtKB_uniProtkbId'])['UniProtKB_uniProtkbId']
+            if this_upid is None:
+                raise ValueError(f"Could not resolve UniProt primary accession for {this_upac}")
```
mtiberti: is this still primary accession?

this is different from the other two calls

mtiberti (changes requested):

mtiberti (approved):

mtiberti (approved):

PR #305, opened 2026-08-26
Convert to polars for RevelDatabase module

emmaqingjie:
- Replaced panda operations with polars in the RevelDatabase module
- Added lazy csv scanning when loading the REVEL chr ddata
- Added polars as a dependency in the setup.py

Comparing res:
- compared pandas and polar outcome
- tested on TP53 which had 159 missense variants
- Made txt file for each of the two outputs and did a diff for the two files, no diff as REVEL scores were identical

Revel time profiling:
REVEL PROFILING SUMMARY
=======================

PANDAS
Runs: [110.383, 106.729, 116.968]
Average: 111.360 s

POLARS
Runs: [8.266, 8.505, 8.599]
Average: 8.457 s

Speedup: 13.17x

See specific numbers below, but profiling shows that pandas spends most of the annotation time in repeated series/string comparisons (aorund 54-59 s cumulative per run). Polars: main bottleneck is the lazyframe collect/chromosome parsing (6-8 s), later df filtering takes around 2 s. 

PANDAS BOTTLENECKS
------------------
2/1    7.550    3.775   69.454   69.454 datasources.py:2736(add_metadata)
1    0.431    0.431   69.454   69.454 datasources.py:2764(_get_revel)
966    0.005    0.000   58.783    0.061 pandas/core/series.py:6726(_cmp_method)
966   17.754    0.018   57.487    0.060 pandas/core/arrays/string_.py:1176(_cmp_method)
1    1.679    1.679    9.140    9.140 datasources.py:2704(_filter_revel_by_chromosomes)
2/1    9.053    4.527   63.502   63.502 datasources.py:2736(add_metadata)
1    0.309    0.309   63.502   63.502 datasources.py:2764(_get_revel)
966    0.005    0.000   54.922    0.057 pandas/core/series.py:6726(_cmp_method)
966   17.694    0.018   53.958    0.056 pandas/core/arrays/string_.py:1176(_cmp_method)
1    0.615    0.615    7.182    7.182 datasources.py:2704(_filter_revel_by_chromosomes)

POLARS BOTTLENECKS
------------------
2/1    0.002    0.001    8.263    8.263 datasources.py:2733(add_metadata)
1    0.022    0.022    8.263    8.263 datasources.py:2761(_get_revel)
1128    6.674    0.006    6.674    0.006 {method 'collect' of 'builtins.PyLazyFrame' objects}
1    0.000    0.000    6.135    6.135 datasources.py:2705(_filter_revel_by_chromosomes)
805    0.006    0.000    1.976    0.002 polars/dataframe/frame.py:5392(filter)
322    0.002    0.000    0.031    0.000 polars/dataframe/frame.py:10588(select_seq)
2/1    0.013    0.007    8.504    8.504 datasources.py:2733(add_metadata)
1128    8.267    0.007    8.267    0.007 {method 'collect' of 'builtins.PyLazyFrame' objects}
1    0.000    0.000    6.353    6.353 datasources.py:2705(_filter_revel_by_chromosomes)
805    0.008    0.000    2.010    0.002 polars/dataframe/frame.py:5392(filter)

Closes #304


cancermuts/datasources.py
```diff
@@ -2704,34 +2705,30 @@ def __init__(self, revel_file):
...
-
-            missing = [c for c in cols if c not in header]
-            if missing:
-                raise ValueError(f"[REVEL] Missing columns in file: {missing}")
-
-            prefix = f"{chrom},"
+        # scan_csv does lazy read, reads data with .collect()
+        lazy_df = pl.scan_csv(
+            file_path,
+            schema_overrides={col: pl.String for col in cols}
```
mtiberti: why do we use this `schema_overrides` here and coerce everything to strings?

e.g. the `_pos` column could be integers, `REVEL` is def going to be a floating point number

cancermuts/datasources.py
```diff
@@ -2704,34 +2705,30 @@ def __init__(self, revel_file):
     def _filter_revel_by_chromosomes(self, chrom):
```
mtiberti: I think we should try and get rid of this mechanism by which we pre-load chromosomes, and probably of this function

this was designed in that way because it was faster than parsing with `pandas`; but if `polars` is as fast as they say, there should be no need to divide the parsing by chromosome, caching etc

the structure can probably be simplified into a single function that does everything

cancermuts/datasources.py
```diff
@@ -2704,34 +2705,30 @@ def __init__(self, revel_file):
...
+            raise ValueError(f"[REVEL] Missing columns in file: {missing}")
 
-        if not filtered_lines:
-            self.log.warning(f"[REVEL] No entries found for chromosomes {chrom}")
-            return pd.DataFrame(columns=cols)
+        df = (lazy_df
+            .select(cols)
+            .filter(pl.col("chr") == str(chrom))
+            .collect()
+            )
```
mtiberti: even if we keep a single function, for now I would still keep this two-step process, in which

1. we filter by chromosome
2. we filter by mutations

it could still be advantageous, performance-wise

cancermuts/datasources.py
```diff
@@ -2776,13 +2773,13 @@ def _get_revel(self, mutation_entries):
...
-                ignore_index=True
+            df = pl.concat(
+                [self._revel_cache_by_chr[c] for c in needed_chroms]
             )
         else:
-            df = pd.DataFrame(columns=["chr", "hg19_pos", "grch38_pos", "ref", "alt",
-                                       "aaref", "aaalt", "REVEL", "Ensembl_transcriptid"])
+            cols = ["chr", "hg19_pos", "grch38_pos", "ref", "alt",
+                    "aaref", "aaalt", "REVEL", "Ensembl_transcriptid"]
+            df = pl.DataFrame(schema={col: pl.String for col in cols})
```
mtiberti: same comment as before re type coercing

cancermuts/datasources.py
```diff
@@ -2806,53 +2803,58 @@ def _get_revel(self, mutation_entries):
                         mutation.metadata['revel_score'].append(Revel(source=self, score=s))
                     continue
 
-                df_filtered = df[df[coord_col].astype(str) == str(gm.get_coord())]
-                df_filtered = df_filtered[df_filtered["chr"].astype(str) == str(gm.chr)]
-                df_filtered = df_filtered[df_filtered["alt"] == gm.alt]
+                df_filtered = df.filter(
+                    (pl.col(coord_col).cast(pl.String) == str(gm.get_coord())) &
+                    (pl.col("chr").cast(pl.String) == str(gm.chr)) &
+                    (pl.col("alt") == gm.alt))
```
mtiberti: do we really need these `.cast` and `str`?

I understand they need to be of the same type to be compared, but we should try and cast `gm.get_coord` to whatever type the dataframe has

this is because I think casting (changing type) of a whole column will be more expensive than changing type of a single value

the two types should be matching but shouldn't necessarily be strings

cancermuts/datasources.py
mtiberti: I think we can get rid of this and just filter for chromsomes directly here, getting rid of the `_revel_cache_by_chr` function

cancermuts/datasources.py
```diff
@@ -2776,13 +2773,13 @@ def _get_revel(self, mutation_entries):
 
         # Build a working DataFrame from just the needed chromosomes
         if needed_chroms:
-            df = pd.concat(
-                [self._revel_cache_by_chr[c] for c in needed_chroms],
-                ignore_index=True
+            df = pl.concat(
+                [self._revel_cache_by_chr[c] for c in needed_chroms]
```
mtiberti: this also won't be needed because we can filter for multiple values at once

mtiberti (changes requested):

cancermuts/datasources.py
emmaqingjie: removed the `schema_overrides `

cancermuts/datasources.py
emmaqingjie: this was removed entirely

cancermuts/datasources.py
emmaqingjie: Not sure I understand this comment. But it should now still follow a two-step process where we filter by chromosome first and then mutations

cancermuts/datasources.py
emmaqingjie: Removed

cancermuts/datasources.py
emmaqingjie: Removed. I filter for all needed chromosomes at once using` is_in(needed_chroms)`.

cancermuts/datasources.py
emmaqingjie: no we don't - removed the `.cast() `calls om df columns. Now I convert individual gm values to match the corresponding df types instead: the coordinate to int, and the chromosome to str since the REVEL chromosome column can also contain values such as 'X'.

emmaqingjie:
Re-ran revel profiling:
Polars before review changes: 8.4 s average
Polars after review changes: 9.7s average
Pandas baseline: 111.4 s average

Reviewd Polars implementation: Rate-limiting step is the LazyFrame collect() / CSV materialization step (~7.2 s cumulative in run 5), while df filtering accounts for ~1.9 s.

cancermuts/datasources.py
```diff
@@ -2769,20 +2737,29 @@ def _get_revel(self, mutation_entries):
...
-            )
-        else:
-            df = pd.DataFrame(columns=["chr", "hg19_pos", "grch38_pos", "ref", "alt",
-                                       "aaref", "aaalt", "REVEL", "Ensembl_transcriptid"])
+        df = (
+            lazy_df
+            .select(cols)
+            .filter(pl.col("chr").is_in(needed_chroms))
+            .collect()
+        )
```
mtiberti: I've done a couple of tests to make this faster

both using the streaming engine and using `==` with single chromosome instead of `is_in` resulted in negligible gains

cancermuts/datasources.py
```diff
@@ -2806,71 +2783,72 @@ def _get_revel(self, mutation_entries):
...
-                    try:
-                        score = float(score_str)
-                        parsed_scores.append(score)
-                        self.log.info(
-                            f"[REVEL] Match for {mutation}: chr={gm.chr}, pos={gm.get_coord()}, "
-                            f"ref={gm.ref}, alt={gm.alt}, aaref={mutation.ref}, "
-                            f"aaalt={mutation.alt}, transcript={transcript_id}, score={score}")
-                    except ValueError:
-                        self.log.warning(f"[REVEL] Could not parse REVEL score '{score_str}' for {mutation}")
+                for score in df_final["REVEL"].drop_nulls():
```
mtiberti: can we add a warning for the values that we find `null`?

mtiberti: this is would be roughly equivalent to the `except` clause in the legacy code

mtiberti (changes requested):
I would also like to try (and time) a different approach

right now we have something like

```
        for mutation, gms, transcript_id in mutation_entries:
                 ...
                df_filtered = df.filter(
                    (pl.col(coord_col) == int(gm.get_coord())) &
                    (pl.col("chr") == str(gm.chr)) &
                    (pl.col("alt") == gm.alt))
```

an alternative to this could be to build a dataframe with all gms and then `join` it (left join) with the main REVEL database:

```
df_mutations = pl.DataFrame(...)
df_mutations.join(revel_df, how='left', on=[...]
```

we could try this in two ways before or after filtering by chromosome. We could find that doing the `join` could be fast enough to not have to filter by chromosome separately as we do, further simplifying our code

notice that to test/time this you don't necessarily need to do it in the "main" code at first, you can use a script at the side that just parses the REVEL database with some mutations or in iPython, but it's up to you

cancermuts/datasources.py
emmaqingjie: Triede benchmarking your three approach suggestions. Did 5 randomized runs using the same 517 TP53 genomic mutation entries.

**Current per-mutation filtering: 13.258 s avg
Chromosome filter + join: 8.217 s avg
Direct join: 8.556 s avg**

All returned 517 matches.
So the join-based approaches are faster than the current set-up. Very little difference in the chromosome filter + joing and direct join. So if we want to keep the code simpler, then direct join could be a good option.

cancermuts/datasources.py
emmaqingjie: added like this
```
for score in df_final["REVEL"]:
                    if score is None:
                        self.log.warning(
                                f"[REVEL] Null REVEL score for {mutation} at chr={gm.chr}, "
                                f"pos={gm.get_coord()}, transcript={transcript_id}"
                                )
                        continue
```

cancermuts/datasources.py
mtiberti: sounds good to me 👍

emmaqingjie:
Time profiling: 20s avg
The direct-join benchmark (~8.6 s) from earlier used inner join and new push uses the suggested mutation-side left join to preserve unmatched mutations - so not directly comparable.

cancermuts/datasources.py
```diff
@@ -2763,31 +2731,93 @@ def add_metadata(self, sequence, md_type=['revel_score']):
...
+                    hasattr(gm, attr)
+                    for attr in ["genome_build", "chr", "get_coord", "ref", "alt"]
+                ):
+                    continue
+
+                if gm.genome_build not in ("hg19", "hg38"):
+                    continue
+
+                mutation_rows.append({
+                    "mutation_idx": mutation_idx,
```
mtiberti: is it necessary to keep this index?

cancermuts/datasources.py
```diff
@@ -2763,31 +2731,93 @@ def add_metadata(self, sequence, md_type=['revel_score']):
...
+                    "gm_idx": gm_idx,
```
mtiberti: is it necessary to keep this index?

PR #306, opened 2026-08-28
 COSMIC datasource to Polars

emmaqingjie:
## Summary
Polars conversion first time around. 
The output was sanity-checked against the Pandas implementation and returns the same 160 mutations for the tested PTPRM transcript. The di

## Performance
Profiling was performed over 5 runs using the same cosmic query:
- Pandas: 109.929 s avg
- Polars: 24.229 s avg
- Approx. 4.5x faster

All 5 Polars runs returned 160 mutations and identical mutation lists.

Closes #290

### DIFF
--- a/cancermuts/datasources.py
+++ b/cancermuts/datasources.py
@@ -35,6 +35,7 @@ from Bio.PDB.Polypeptide import three_to_index, index_to_one
 from Bio import SeqIO, Seq
 import numpy as np
 import pandas as pd
+import polars as pl 
 from .core import Sequence, ProteinVariant
 from .properties import *
 from .metadata import *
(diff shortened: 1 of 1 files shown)


cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
...
+            )
+
             try:
-                df_tmp = pd.read_csv(file, sep='\\t', dtype='str', na_values='NS', usecols=self._use_cols_database_files, encoding=self._encoding)
-                df_tmp['TRANSCRIPT_ACCESSION'] = df_tmp['TRANSCRIPT_ACCESSION'].map(lambda x: x.split('.', 1)[0] if isinstance(x, str) else x)
+                df_tmp = pl.scan_csv(
+                    file,
+                    separator="\t",
+                    null_values="NS",
+                    encoding=self._encoding if self._encoding is not None else "utf8",
```
mtiberti: why did you set this default?

if we need to set a default as this it would be better to set it as the default of the constructor argument and just pass it down, without handling the case here

cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
...
+                separator="\t",
+                null_values="NS",
+                columns=self._use_cols_classification_files,
+                encoding=self._encoding if self._encoding is not None else "utf8",
+            )
+
+        except Exception:
+            self.log.error(
+                f"Couldn't parse database file {classification_db_filename}"
+            )
```
mtiberti: can we have more specific exception(s) than this?

in general catching `Exception` is not a very good practice because it catches anything, preventing us from understanding what is going on

cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
...
+            raise TypeError(
+                f"Couldn't parse database file {classification_db_filename}"
+            )
```
mtiberti: same here - this is not really handling the exception, this is raising it again with a different type - at this point we would be best passing forward the exception that was generated if we don't know what it is

cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
...
+
+        except Exception:
+            self.log.error(
+                "Couldn't merge database files due to missing or incorrectly "
+                "named join columns"
+            )
+            raise TypeError(
+                "Couldn't merge database files due to missing or incorrectly "
+                "named join columns"
+            )
```
mtiberti: same as previous exception

cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
         else:
             self._df = None
 
-    def _load_db_files(self, targeted_db_file, screenmut_db_file):
+    def _load_db_files(self, targeted_db_file, screenmut_db_file, gene_id=None):
```
mtiberti: can we call the `gene_id` argument something different, e.g. `gene_symbol`? This is supposed to be the HUGO name (symbol) of the gene

cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
...
-        tmp_targeted_screenmut_df = pd.concat(targeted_screenmut_dataframes, ignore_index=True, sort=False)
+            except Exception:
+                self.log.error(
+                    f"Couldn't parse database file "
+                    f"{targeted_screenmut_db_filenames[fi]}"
+                )
+                raise TypeError(
+                    f"Couldn't parse database file "
+                    f"{targeted_screenmut_db_filenames[fi]}"
+                )
```
mtiberti: same as previous exceptions

cancermuts/datasources.py
```diff
@@ -1956,58 +2022,85 @@ def _parse_db_files(self, gene_id, transcript_accession, genome_assembly_version
...
+                self._targeted_database_file,
+                self._screen_mutant_database_file,
+                gene_id=gene_id
+            )
 
-            filtered_targeted_database_file = StringIO("".join(filtered_lines_t))
-            filtered_screenmut_database_file = StringIO("".join(filtered_lines_s))
+            if df.is_empty():
+                raise ValueError(
+                    f"The given gene_id {gene_id} is not present in the database files"
```
mtiberti: again, this is not a gene ID - to be revised

cancermuts/datasources.py
mtiberti: Besides, I'm not sure why you prefer to run filtering for gene here instead than in the calling function

cancermuts/datasources.py
```diff
@@ -1884,37 +1885,102 @@ def __init__(self, targeted_database_file, screen_mutant_database_file, classifi
...
+                    )
+
+                df_tmp = (
+                    df_tmp
+                    .with_columns(
+                        pl.col("TRANSCRIPT_ACCESSION")
+                        .str.split(".")
+                        .list.first()
+                    )
+                    .collect()
```
mtiberti: here we have two operating modes:

  - `lazy_load_db=False` in which we want to just load the database dataframes and then use them. In this case, it's fine to `.collect()` early since we will need to work on the dataframes anyway
  - `lazy_load_db=True` in which we want to `.collect()` _as late as possible_ after all the filtering has bene decided, so we can take advantage of polars' optimized evaluation. I don't know if it will make a practical difference in terms of execution time for this case, but it's a general design consideration

so I would distinguish the two cases:
  - this function should return a `LazyFrame` or a `DataFrame` depending on what we want
  - the `_parse_db_files()` function should `.collect()` if necessary, and as late as possible (meaning after all the filtering etc has been done)

mtiberti (changes requested):

mtiberti (changes requested):
also missing changes in CHANGELOG.md

